PR #1, opened 2016-07-12
Add .gitignore for Spark site repo

srowen:
(and test git machinery for new site repo)


rxin:
LGTM!

rxin:
We need to submit a merge script, don't we?

srowen:
Yeah I went ahead and pushed this directly upstream. Works fine if one of us are doing it. If we need to merge others' PRs with any regularity... yeah may want a merge script. The existing one may be literally callable as-is from where it sits.

PR #5, opened 2016-07-19
Fix main links to streaming examples

isaacl:
Lots of other links in documentation are wrong but might be fixed with auto-generated docs.

Could use these:

```
sed -i 's#spark/streaming/examples#spark/examples/streaming#g'
sed -i 's#github.com/apache/incubator-spark#github.com/apache/spark#g'
```

### DIFF
--- a/site/streaming/index.html
+++ b/site/streaming/index.html
@@ -307,7 +307,7 @@
     <ul class="list-narrow">
       <li><a href="/downloads.html">Download Spark</a>. It includes Streaming as a module.</li>
       <li>Read the <a href="/docs/latest/streaming-programming-guide.html">Spark Streaming programming guide</a>, which includes a tutorial and describes system architecture, configuration and high availability.</li>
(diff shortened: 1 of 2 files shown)


srowen:
Thanks @isaacl -- actually the HTML needs to be regenerated too with `jekyll build` in order to reflect the change in the site too. You're welcome to add that, or I can do so separately if you don't care.

isaacl:
Done. 

By the way, jekyll also generated a [diff](https://gist.githubusercontent.com/anonymous/7bccaeeec81940906ba10782c2168f9d/raw) in site/news/index.html. No sure why -- issue with timezones?

srowen:
Yeah I just before 'fixed' a bunch of spurious diffs between jekyll output and the HTML. That one still puzzles me, because the source and file name says 5/16 (fixed from 5/17 before) but for me jekyll has at times rendered 5/15. No idea. I ignored it.

PR #13, opened 2016-10-25
Link to archive.apache.org for Spark releases besides 2.0.1 and 1.6.2

srowen:
Per http://www.apache.org/dev/release#when-to-archive we should periodically remove non-current releases from dist.apache.org. They remain available at archive.apache.org. This change links to dist only for 1.6.2 and 2.0.1 (and so other releases can subsequently be removed from dist).

It also removes 1.3.x from the list of advertised downloads, removes / comments some dead JS in this file and fixes an HTML error. I don't feel strongly about the 1.3 change, just seems like it's useful to keep this list pared down to pretty recent releases, despite the fact that they're always available on archive.apache.org.

### DIFF
--- a/js/downloads.js
+++ b/js/downloads.js
@@ -3,8 +3,8 @@
 
 releases = {};
 
-function addRelease(version, releaseDate, packages, downloadable, stable) {
-  releases[version] = {released: releaseDate, packages: packages, downloadable: downloadable, stable: stable};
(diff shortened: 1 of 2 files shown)


rxin:
LGTM

srowen:
Merged and removed the non-current builds from dist: https://dist.apache.org/repos/dist/release/spark/

PR #16, opened 2016-11-16
Expand guidance on SO and mailing lists, per discussion

srowen:


dennyglee:
Could you add `<a name="have-questions"></a>` so we can link to that section of the web page directly?   Should `<a name="mailing-lists">` be removed since it is a subheading (`<h4>`) instead of a heading (`<h3>`) per se?

srowen:
I can add the anchor, yes. If I remove the mailing-lists anchor then existing links to it won't jump to the heading, so I figured it's worth keeping.

dennyglee:
Oh good point!  Thanks!

PR #18, opened 2016-11-19
(Again again) make web site HTML consistent with latest jekyll output

srowen:
(I need this to stop changing back because i'm going to make a number of changes to port the wiki over, and these diffs will keep creating noise in the pull requests.)


site/mailing-lists.html
```diff
@@ -12,7 +12,7 @@
 
   
     <meta http-equiv="refresh" content="0; url=/community.html">
-    <link rel="canonical" href="http://localhost:4000community.html" />
+    <link rel="canonical" href="/community.html" />
```
srowen, line 15: ... hmm, this snuck in my last commit and is wrong. This is a fix.

PR #20, opened 2016-11-19
remove broken links from the documentation site

cko:
The documentation site contains links to blog posts which no longer exists.


srowen:
Sounds good, but can you run `jekyll build` to generate the actual HTML? and then add the changed HTML to this PR?

cko:
[x] done
What about either adding a Pull Request template or creating a Travis CI job that generates the site?
Then you wouldn't have to ping everyone personally ;-)

srowen:
Yes, I agree with your PR to add some instructions about this.
I don't think we want to auto-generate the site from markdown because it goes live as-is. I think we'd want to review the HTML change as part of the change.

cko:
fine for me

srowen:
BTW I did merge this but the sync seems to be offline again between ASF and github. The problem is that it's hard to push to apache when I can't get the commits present in the ASF repo, because the push will be rejected. Seeing if I can work around this now ...

PR #23, opened 2016-11-19
Update sitemap and make it partly auto-generate; fix site URL

srowen:
This mostly overhauls and updates the sitemap.xml file, which hasn't been updated in almost 2 years, to be more automatic


_config.yml
```diff
@@ -6,11 +6,4 @@ permalink: none
 destination: site
 exclude: ['README.md','content']
 keep_files: ['docs']
-
-# The recommended way of viewing the website on your local machine is via jekyll using
```
srowen, line 10: This info is out of date. `jekyll serve` works as intended without modifying this, and the site URL should be the full host / path

PR #24, opened 2016-11-22
Add chat rooms to community links

jodersky:
We've had quite some activity on the spark-with-scala Gitter channel, although it was only advertised once in an email thread. Adding it to the community page would help in making it more visible.

I went ahead and put together a template for links to chat rooms, let me know if there are any others and I'll add them to the list.

### DIFF
--- a/community.md
+++ b/community.md
@@ -72,6 +72,16 @@ Some quick tips when using email:
 and include only a few lines of the pertinent code / log within the email.
 - No jobs, sales, or solicitation is permitted on the Apache Spark mailing lists.
 
+<a name="chat"></a>
+<h4>Chat Rooms</h4>
+
+Chat rooms are great for quick questions or discussions on specialized topics. The following chat rooms are not officially part of Apache Spark; they are provided for reference only.
+<ul>
+  <li>
+    <a href="https://gitter.im/spark-scala/Lobby">Spark with Scala</a> is for questions and discussions related to using Spark with the Scala programming language.
+  </li>
+</ul>
+
 <a name="events"></a>
 <h3>Events and Meetups</h3>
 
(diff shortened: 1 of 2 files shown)


deanwampler:
+1

community.md
```diff
@@ -72,6 +72,16 @@ Some quick tips when using email:
 and include only a few lines of the pertinent code / log within the email.
 - No jobs, sales, or solicitation is permitted on the Apache Spark mailing lists.
 
+<a name="chat"></a>
+<h4>Chat Rooms</h4>
+
+Chat rooms are great for quick questions or discussions on specialized topics. The following chat rooms are not officially part of Apache Spark; they are provided for reference only.
+<ul>
+  <li>
+    <a href="https://gitter.im/spark-scala/Lobby">Spark with Scala</a> is for questions and discussions related to using Spark with the Scala programming language.
```
srowen, line 81: Yes, I think linking to 'unofficial' chats is OK if they're active and useful and any such resource is linked. 

I have no idea what the theory is about HTML vs markdown but this can use simple `- ...` syntax because the other docs do. I don't feel strongly. HTML is used inconsistently.

I will probably want to clean up the Community menu along with some other changes here because it's getting busy and "Mailing Lists" no longer described what is really linked here. I'll do that separately.

jodersky: > no idea what the theory is about HTML vs markdown but this can use simple -

agree, I somehow missed that some parts of the document actually used markdown (html drowns the content in tags), hence to stay consistent with the rest I used html.

Should I change this or would you like to do it in the clean up you mention?

srowen:
Merged, though I don't think it will auto-close because I had to rebase before pushing. Can you close @jodersky ?

SethTisue:
fwiw there's also the #apache-spark IRC room

PR #33, opened 2017-02-12
Move information about Zinc and running individual tests to main website

kayousterhout:
This moves the information about how to run individual tests and use Zinc to the developer tools page, from the release-specific docs.  I've found the developer tools page is more Google-able, this info is useful primarily for developers, and it doesn't change significantly between releases.   There's a companion PR in the main Spark repo to remove this from the release-specific docs: https://github.com/apache/spark/pull/17018

### DIFF
--- a/developer-tools.md
+++ b/developer-tools.md
@@ -9,7 +9,9 @@ navigation:
 
 <h2>Useful Developer Tools</h2>
 
-<h3>Reducing Build Times</h3>
+<h3 id="reducing-build-times">Reducing Build Times</h3>
+
+<h4>SBT: Avoiding Re-Creating the Assembly JAR</h4>
 
 Spark's default build strategy is to assemble a jar including all of its dependencies. This can 
 be cumbersome when doing iterative development. When developing locally, it is possible to create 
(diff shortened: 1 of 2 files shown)


srowen:
I didn't copy this from the wiki because it had been added earlier to the docs at http://spark.apache.org/docs/latest/building-spark.html#running-tests  That doesn't mean it's the best place for this piece of info. If you'd like to add this then just also make a PR to remove similar info in the release-specific docs (perhaps with a pointer to this new info)

kayousterhout:
Ah thanks for the pointer to that -- I hadn't seen that page.  It looks like the Building Spark page has some stuff that's really only for developers (mainly the Zinc info and how to run individual tests).  What do you think about me moving all of that (so the Zinc info in addition to the individual tests info) to the developers page, and then adding a link on the building spark page to the developer page for more info on running tests?

If I do that, do you think I should move *all* of the "testing" info to the developer page? I'm inclined to think no because I can imagine non-developers wanting to run all of the tests (e.g., to verify a release), but figured you'd have better visibility into this, so wanted to check.

srowen:
Feel free to move whatever you think should be moved. When I merged the wiki into the main site it was a rough cut at fitting bits together that weren't necessarily in the best place, so more editing is certainly welcome. 

I can help review and merge this and/or the complementary change to the main docs whenever you're ready.

srowen:
@kayousterhout are you interested in updating the main project docs too? we can commit both changes.

kayousterhout:
Yup sorry for the delay on this -- was out on vacation.  I'll do both today.

kayousterhout:
Ok I  updated this to add the Zinc stuff, and also added a corresponding PR in the main Spark repo.  We should merge this one first so the links I added in the Spark release-specific docs work.

srowen:
@kayousterhout this is merged now though it won't auto-close because I rebased/squashed

kayousterhout:
Cool thanks for the help with this Sean

PR #34, opened 2017-02-14
[SPARK-19550][SPARK-19464] Remove references to Java 7 and Hadoop support before 2.5

srowen:
To go with SPARK-19550, removed references to Java 7.
And, this removes references to Hadoop 2.5 and earlier for SPARK-19464.

Regenerating the site, I again again again see differences in output due to people committing the output of old versions of `jekyll`. Please update!


srowen:
I'm going to merge this now, even though this reflects changes in 2.2.0+ which isn't released yet, because the changes don't remove any information that's important for 2.1, nor add info that would only make sense for 2.2.

(Most of the change is again fixing the output from others jekylls too)

PR #42, opened 2017-03-25
Add section about mima in developer tools.

BenFradet:
I think this would be useful for new contributors that might not be familiar with what MiMa is.

### DIFF
--- a/developer-tools.md
+++ b/developer-tools.md
@@ -111,6 +111,60 @@ To run individual Java tests, you can use the `-Dtest` flag:
 build/mvn test -DwildcardSuites=none -Dtest=org.apache.spark.streaming.JavaAPISuite test
 ```
 
+<h3>Binary compatibility</h3>
+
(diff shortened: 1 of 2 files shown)


developer-tools.md
```diff
@@ -111,6 +111,58 @@ To run individual Java tests, you can use the `-Dtest` flag:
...
+Test build #xx has finished for PR yy at commit ffffff.
+
+  This patch fails MiMa tests.
+  This patch merges cleanly.
+  This patch adds no public classes.
+```
+
+<h4>Solving a binary incompatibility</h4>
+
+If you believe that your binary incompatibilies are justified, you can filter
```
srowen: Jira -> JIRA
Also it's more frequent that the incompatibility is a 'false positive' because it doesn't affect the user API. It's not always that it's a break but it's justified.

srowen (approved):
Tiny comment but I don't feel strongly about it. This seems fine to add.

BenFradet:
@srowen done :+1:

srowen (approved):

PR #45, opened 2017-04-13
add intellij information

samelamin:
Updated HTML as requested by @srowen


developer-tools.md
```diff
@@ -292,6 +292,7 @@ so, open the "Project Settings" and select "Modules". Based on your selected Mav
 may need to add source folders to the following modules:
     - spark-hive: add v0.13.1/src/main/scala
     - spark-streaming-flume-sink: add target\scala-2.10\src_managed\main\compiled_avro
+    - spark-catalyst: add target\scala-2.11\src_managed\main
```
samelamin: should I change this to scala-2.10? I think we should change the documentation to point at 2.11 since 2.10 is in the process of being deprecated

srowen:
OK to point at 2.11. read the note above -- you need to regenerate the HTML too.

PR #65, opened 2017-09-11
Update YourKit usage in developer tools

maropu:

### DIFF
--- a/developer-tools.md
+++ b/developer-tools.md
@@ -394,20 +394,19 @@ consider mirroring this file or including it on a custom AMI.
 - Configure the Spark JVMs to use the YourKit profiling agent by editing `~/spark/conf/spark-env.sh` 
 and adding the lines
 ```
-SPARK_DAEMON_JAVA_OPTS+=" -agentpath:/root/YourKit-JavaProfiler-2017.02/bin/linux-x86-64/libyjpagent.so=sampling,port=10001"
+SPARK_DAEMON_JAVA_OPTS+=" -agentpath:/root/YourKit-JavaProfiler-2017.02/bin/linux-x86-64/libyjpagent.so=sampling"
 export SPARK_DAEMON_JAVA_OPTS
 ```
 - Copy the updated configuration to each node: `~/spark-ec2/copy-dir ~/spark/conf/spark-env.sh`
 - Restart your Spark cluster: `~/spark/bin/stop-all.sh` and `~/spark/bin/start-all.sh`
-- By default, the YourKit profiler agents use ports `10001-10010`. In the example above,
-we explicitly use `10001` for the agents. To connect the YourKit desktop
+- By default, the YourKit profiler agents use ports `10001-10010`. To connect the YourKit desktop
 application to the remote profiler agents, you'll have to open these ports in the cluster's EC2 
 security groups. To do this, sign into the AWS Management Console. Go to the EC2 section and 
 select `Security Groups` from the `Network & Security` section on the left side of the page. 
 Find the security groups corresponding to your cluster; if you launched a cluster named `test_cluster`, 
 then you will want to modify the settings for the `test_cluster-slaves` and `test_cluster-master` 
 security groups. For each group, select it from the list, click the `Inbound` tab, and create a 
-new `Custom TCP Rule` opening the port range `10001` in the example. Finally, click `Apply Rule Changes`.
+new `Custom TCP Rule` opening the port range `10001-10010`. Finally, click `Apply Rule Changes`. 
 Make sure to do this for both security groups.
 Note: by default, `spark-ec2` re-uses security groups: if you stop this cluster and launch another 
 cluster with the same name, your security group settings will be re-used.
--- a/developer-tools.md
+++ b/developer-tools.md
@@ -389,26 +389,25 @@ Here are instructions on profiling Spark applications using YourKit Java Profile
 <a href="https://www.yourkit.com/download/index.jsp">YourKit downloads page</a>. 
 This file is pretty big (~100 MB) and YourKit downloads site is somewhat slow, so you may 
 consider mirroring this file or including it on a custom AMI.
-- Untar this file somewhere (in `/root` in our case): `tar xvjf yjp-12.0.5-linux.tar.bz2`
-- Copy the expanded YourKit files to each node using copy-dir: `~/spark-ec2/copy-dir /root/yjp-12.0.5`
+- Unzip this file somewhere (in `/root` in our case): `unzip YourKit-JavaProfiler-2017.02-b66.zip`
+- Copy the expanded YourKit files to each node using copy-dir: `~/spark-ec2/copy-dir /root/YourKit-JavaProfiler-2017.02`
 - Configure the Spark JVMs to use the YourKit profiling agent by editing `~/spark/conf/spark-env.sh` 
 and adding the lines
 ```
-SPARK_DAEMON_JAVA_OPTS+=" -agentpath:/root/yjp-12.0.5/bin/linux-x86-64/libyjpagent.so=sampling"
+SPARK_DAEMON_JAVA_OPTS+=" -agentpath:/root/YourKit-JavaProfiler-2017.02/bin/linux-x86-64/libyjpagent.so=sampling,port=10001"
 export SPARK_DAEMON_JAVA_OPTS
-SPARK_JAVA_OPTS+=" -agentpath:/root/yjp-12.0.5/bin/linux-x86-64/libyjpagent.so=sampling"
-export SPARK_JAVA_OPTS
 ```
 - Copy the updated configuration to each node: `~/spark-ec2/copy-dir ~/spark/conf/spark-env.sh`
 - Restart your Spark cluster: `~/spark/bin/stop-all.sh` and `~/spark/bin/start-all.sh`
-- By default, the YourKit profiler agents use ports 10001-10010. To connect the YourKit desktop 
+- By default, the YourKit profiler agents use ports `10001-10010`. In the example above,
+we explicitly use `10001` for the agents. To connect the YourKit desktop
 application to the remote profiler agents, you'll have to open these ports in the cluster's EC2 
 security groups. To do this, sign into the AWS Management Console. Go to the EC2 section and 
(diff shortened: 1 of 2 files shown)


developer-tools.md
```diff
@@ -389,26 +389,25 @@ Here are instructions on profiling Spark applications using YourKit Java Profile
 <a href="https://www.yourkit.com/download/index.jsp">YourKit downloads page</a>. 
 This file is pretty big (~100 MB) and YourKit downloads site is somewhat slow, so you may 
 consider mirroring this file or including it on a custom AMI.
-- Untar this file somewhere (in `/root` in our case): `tar xvjf yjp-12.0.5-linux.tar.bz2`
-- Copy the expanded YourKit files to each node using copy-dir: `~/spark-ec2/copy-dir /root/yjp-12.0.5`
+- Unzip this file somewhere (in `/root` in our case): `unzip YourKit-JavaProfiler-2017.02-b66.zip`
+- Copy the expanded YourKit files to each node using copy-dir: `~/spark-ec2/copy-dir /root/YourKit-JavaProfiler-2017.02`
```
maropu, line 393: Could we still use `spark-ec2/copy-dir`? (not sure to keep `spark-ec2` updated)

maropu:
@srowen could you check? thanks!

developer-tools.md
```diff
@@ -389,26 +389,25 @@ Here are instructions on profiling Spark applications using YourKit Java Profile
...
 - Configure the Spark JVMs to use the YourKit profiling agent by editing `~/spark/conf/spark-env.sh` 
 and adding the lines
 ```
-SPARK_DAEMON_JAVA_OPTS+=" -agentpath:/root/yjp-12.0.5/bin/linux-x86-64/libyjpagent.so=sampling"
+SPARK_DAEMON_JAVA_OPTS+=" -agentpath:/root/YourKit-JavaProfiler-2017.02/bin/linux-x86-64/libyjpagent.so=sampling,port=10001"
```
HyukjinKwon: I thought `port` in the agent can actually omitted because it actually allows omitting the port (detecting the ports `10001-10010`) when YourKit in the desktop connects to the remote applications in the agents.

HyukjinKwon (commented):
Look fine to me except a question. Would you mind regenerating HTML by `jekyll build` and include the changes too? This will make some diff, for example in `site/developer-tools.html`.

developer-tools.md
maropu: yea, I a little misunderstood this. I'll update soon.

maropu:
@HyukjinKwon Thanks! yea, sure. I'll add gen'd html, too.

maropu:
@HyukjinKwon I only added diffs in `site/developer-tools.html` though, your original intention was to add other diffs in `/site`? I just run `jekyll build` and got other diffs below;
```
        modified:   site/committers.html
        modified:   site/contributing.html
        modified:   site/downloads.html
        modified:   site/examples.html
        modified:   site/news/index.html
        modified:   site/release-process.html
        modified:   site/security.html
        modified:   site/sitemap.xml
```

HyukjinKwon:
Ah, sure. That's the same case to me as well. I am not aware of why in details but I just think different versions might affect how to generate HTMLs and/or some custom changes were made to the HTMLs.
In any event, I manually filtered out unrelated changes when I proposed few PRs before.

maropu:
oh, yea. ok. Even in this pr, is it okay to filter out unnecessary diffs manually?

HyukjinKwon:
Yes, I think that should be safer.

maropu:
ok, I'll update.

maropu:
ok, updated

site/developer-tools.html
```diff
@@ -568,19 +568,17 @@ <h3 id="running-individual-tests">Running Individual Tests</h3>
 <a href="https://www.yourkit.com/download/index.jsp">YourKit downloads page</a>. 
 This file is pretty big (~100 MB) and YourKit downloads site is somewhat slow, so you may 
 consider mirroring this file or including it on a custom AMI.</li>
-  <li>Untar this file somewhere (in <code>/root</code> in our case): <code>tar xvjf yjp-12.0.5-linux.tar.bz2</code></li>
-  <li>Copy the expanded YourKit files to each node using copy-dir: <code>~/spark-ec2/copy-dir /root/yjp-12.0.5</code></li>
+  <li>Unzip this file somewhere (in <code>/root</code> in our case): <code>unzip YourKit-JavaProfiler-2017.02-b66.zip</code></li>
+  <li>Copy the expanded YourKit files to each node using copy-dir: <code>~/spark-ec2/copy-dir /root/YourKit-JavaProfiler-2017.02</code></li>
   <li>Configure the Spark JVMs to use the YourKit profiling agent by editing <code>~/spark/conf/spark-env.sh</code> 
 and adding the lines
-    <pre><code>SPARK_DAEMON_JAVA_OPTS+=" -agentpath:/root/yjp-12.0.5/bin/linux-x86-64/libyjpagent.so=sampling"
```
HyukjinKwon, line 575: @maropu, it looks we should keep `<pre>`. I just double checked the format is broken:

HyukjinKwon (approved):
I at least manually tested and verified YourKit (not ec2 but it looks correct). It looks fine to me except the minor comment.

site/developer-tools.html
maropu: ok, will recheck

developer-tools.md
HyukjinKwon: Hi @shivaram, if I understood correctly, `amplab/spark-ec2` is managed by you. Would you mind if I ask whether we should keep this here? The change itself looks fine if I haven't missed something but I am less sure if we should keep this here or not.

maropu:
I also checked if this worked well in ec2.

developer-tools.md
```diff
@@ -389,19 +389,17 @@ Here are instructions on profiling Spark applications using YourKit Java Profile
...
+- Unzip this file somewhere (in `/root` in our case): `unzip YourKit-JavaProfiler-2017.02-b66.zip`
+- Copy the expanded YourKit files to each node using copy-dir: `~/spark-ec2/copy-dir /root/YourKit-JavaProfiler-2017.02`
 - Configure the Spark JVMs to use the YourKit profiling agent by editing `~/spark/conf/spark-env.sh` 
 and adding the lines
 ```
-SPARK_DAEMON_JAVA_OPTS+=" -agentpath:/root/yjp-12.0.5/bin/linux-x86-64/libyjpagent.so=sampling"
+SPARK_DAEMON_JAVA_OPTS+=" -agentpath:/root/YourKit-JavaProfiler-2017.02/bin/linux-x86-64/libyjpagent.so=sampling"
 export SPARK_DAEMON_JAVA_OPTS
-SPARK_JAVA_OPTS+=" -agentpath:/root/yjp-12.0.5/bin/linux-x86-64/libyjpagent.so=sampling"
-export SPARK_JAVA_OPTS
```
HyukjinKwon, line 400: Wait .. is it okay to remove this out without setting other alternatives (e.g., in cluster mode)?

maropu: I see and I didn't check. probably, is it better that we use `spark.executor.extraJavaOptions` and `spark.driver.extraJavaOptions` for this example? I think these options seem to work even in a cluster mode?

HyukjinKwon: The doc says `SPARK_DAEMON_JAVA_OPTS` for master and worker. Probably replacing it to `spark.driver.extraJavaOptions` should be fine but .. this isn't really my area ..

maropu: I checked the previous pr (https://github.com/apache/spark/pull/17212) to remove `SPARK_JAVA_OPTS` and the removed warning messages suggested we should use instead:
```
          |SPARK_JAVA_OPTS was detected (set to '$value').
          |This is deprecated in Spark 1.0+.
          |
          |Please instead use:
          | - ./spark-submit with conf/spark-defaults.conf to set defaults for an application
          | - ./spark-submit with --driver-java-options to set -X options for a driver
          | - spark.executor.extraJavaOptions to set -X options for executors
          | - SPARK_DAEMON_JAVA_OPTS to set java options for standalone daemons (master or worker)
```
From this message, `SPARK_DAEMON_JAVA_OPTS` seems to work in standalone only... I need to check more..

HyukjinKwon: Yes, I think this isn't my area .. (I am less sure if it'd work in Yarn and etc.). Probably the safe choice should be to set both ..

maropu: yea, ok. It's okay that we just wait for other qualified developer's comments :)

maropu:
I checked the related code and I found `SPARK_EXECUTOR_OPTS` in [SparkClassCommandBuilder](https://github.com/apache/spark/blob/master/launcher/src/main/java/org/apache/spark/launcher/SparkClassCommandBuilder.java#L52). This option seems to be passed into executors for `CoarseGrainedExecutorBackend `(standalone/yarn) and `MesosExecutorBackend `(mesos). 

Btw, I noticed that this option is not described in [conf/spark-env.sh.template](https://github.com/apache/spark/blob/master/conf/spark-env.sh.template)...

developer-tools.md
shivaram: I think its probably good to add some wording on top saying `These instructions apply to when Spark is run using the spark-ec2[link to amplab/spark-ec2] scripts`

maropu: Since this yourkit usage is not only for `spark-ec2`, I think the instruction might make others misunderstood. WDYT? @HyukjinKwon

HyukjinKwon: Actually, that's what I was thinking too. But I would rather avoid to fix it here now as I guess it maybe needs to wait for some more opinions / discussion (e.g., other profiles tools, description for copying it to other nodes, checking duplicated documentation and etc.). I am fine with it if any committer strongly prefers (I am +0).

Otherwise, let's push what we are sure of first as the current changes look obviously something to be fixed. I don't want to make this PR complicated for now and the original intention looks describing yourkit with ec2 anyway.

srowen:
Pushed to the site, @maropu . You'll have to close this manually as I squashed the commits

maropu:
ok, thanks, Sean!

PR #75, opened 2017-11-15
[SPARK-22525] Fix web download javascript

holdenk:
Right now we don't process the package selection (e.g. which version of hadoop the user wants to download for). Restore the `onPackageSelect` function.

### DIFF
--- a/js/downloads.js
+++ b/js/downloads.js
@@ -99,6 +99,14 @@ function initReleaseNotes() {
   }
 }
 
+function onPackageSelect() {
+  var packageSelect = document.getElementById("sparkPackageSelect");
+
+  var pkg = getSelectedValue(packageSelect);
+
+  updateDownloadLink();
+}
+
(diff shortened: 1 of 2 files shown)


holdenk:
cc @srowen @Pashugan

holdenk:
Merged to asf-site

js/downloads.js
```diff
@@ -99,6 +99,14 @@ function initReleaseNotes() {
   }
 }
 
+function onPackageSelect() {
+  var packageSelect = document.getElementById("sparkPackageSelect");
```
srowen, line 103: Oh I missed this; I opened https://github.com/apache/spark-website/pull/76
Only the last line in this method is needed but it won't matter

srowen:
For the record I only managed to commit half of this change -- I rebased incorrectly or something. I'll fix this for real shortly.

PR #80, opened 2017-12-15
Fix the link of 'Mesos Getting Started page'

mcavdar:


srowen:
@mcavdar we usually don't patch the docs for a release, as it would make them inconsistent with the released software artifact. This kind of thing would just get fixed with the next release. You can close this.

mcavdar:
@srowen reopened to fix urls in [documentation](https://spark.apache.org/documentation.html) and [release post](https://spark.apache.org/releases/spark-release-0-9-0.html)

site/docs/2.2.1/running-on-mesos.html
```diff
@@ -198,7 +198,7 @@ <h2 id="from-source">From Source</h2>
 <ol>
   <li>Download a Mesos release from a
 <a href="http://www.apache.org/dyn/closer.lua/mesos/1.0.0/">mirror</a></li>
-  <li>Follow the Mesos <a href="http://mesos.apache.org/gettingstarted">Getting Started</a> page for compiling and
+  <li>Follow the Mesos <a href="http://mesos.apache.org/getting-started">Getting Started</a> page for compiling and
```
srowen, line 201: I still wouldn't change this as it makes the docs inconsistent with the released artifact, even though it did contain a bug. We'd probably only patch this in case of a serious error.

PR #83, opened 2018-01-03
Change company to Oath from Yahoo

tgravescs:
I ran jekyll build and it looks like it reordered the sitemap.  Let me know if that is an issue.

otherwise update the company from Yahoo to Oath.

### DIFF
--- a/committers.md
+++ b/committers.md
@@ -20,10 +20,10 @@ navigation:
 |Ankur Dave|UC Berkeley|
 |Aaron Davidson|Databricks|
 |Thomas Dudziak|Facebook|
-|Robert Evans|Yahoo!|
+|Robert Evans|Oath|
 |Wenchen Fan|Databricks|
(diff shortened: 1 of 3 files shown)


srowen:
Yeah I think several people are running different jekyll versions or plugin versions that don't produce the same sitemap. I usually just revert that part rather than keep changing it, but it doesn't matter.

Just for my curiosity ... I thought Oath was just an umbrella company. The engineering and so on is also under that company not the original Yahoo!?

Looks fine.

tgravescs:
Oath is the actual company now (owned by Verizon), its basically combined Yahoo and aol, Yahoo is just a brand under that company.  Either way I can revert that file if you want or just leave it

srowen:
Done. You'll have to close this manually because I rebased on the asf-site branch.

tgravescs:
thanks, closing.

PR #85, opened 2018-01-08
Update remaining refs to 2.2.0 as latest Spark, to 2.2.1

srowen:


srowen:
CC @felixcheung

documentation.md
```diff
@@ -12,6 +12,7 @@ navigation:
 <p>Setup instructions, programming guides, and other documentation are available for each stable version of Spark below:</p>
 
 <ul>
+  <li><a href="{{site.baseurl}}/docs/2.2.1/">Spark 2.2.1</a></li>
```
felixcheung, line 15: this was never listing the latest for some reasons
I was documented it as it was, but we could change this https://github.com/apache/spark-website/blame/asf-site/release-process.md#L241

site/downloads.html
```diff
@@ -242,7 +242,7 @@ <h3 id="link-with-spark">Link with Spark</h3>
 
 <pre><code>groupId: org.apache.spark
 artifactId: spark-core_2.11
-version: 2.2.0
+version: 2.2.1
```
felixcheung, line 245: I found one more on line https://github.com/apache/spark-website/pull/85/files#diff-6f722d4569121679e85d607509a56baaR257

felixcheung (approved):
thanks, I noticed this earlier and was half way through.
I think one version string remains https://github.com/apache/spark-website/pull/85/files#diff-2a4af035d9e12b64c084b5e7cfb2c420R62

documentation.md
srowen: You are right; I figured add it for completeness rather than consistency. I think the release docs are close enough still as they indicate this needs to change and it's clear how.

srowen:
I thought the comment about https://github.com/apache/spark-website/pull/85/files#diff-2a4af035d9e12b64c084b5e7cfb2c420R62 was actually right as "2.2.0" because I guess it's a branch starting with 2.2.0 (roughly) with maintenance fixes on top. 2.2.1 seemed less accurate.

felixcheung:
ok fair enough

felixcheung (approved):
LGTM

PR #86, opened 2018-01-10
Note update of Jenkins jobs as part of cutting new release branch

srowen:


release-process.md
```diff
@@ -39,8 +39,18 @@ If you are a new Release Manager, you can read up on the process from the follow
...
+It is also good to set up Jenkins jobs for the release branch once it is cut to
+ensure tests are passing. These are jobs like 
+https://amplab.cs.berkeley.edu/jenkins/view/Spark%20QA%20Test/job/spark-branch-2.3-test-maven-hadoop-2.7/ .
+Consult Josh Rosen and Shane Knapp for help with this. Also remember to add the newly-added jobs
+to the test dashboard at https://amplab.cs.berkeley.edu/jenkins/view/Spark%20QA%20Test%20(Dashboard)/ .
+
+Finally, after a new branch is cut, update SNAPSHOT versions found in the `master` branch to the
+next minor version. For example when cutting `branch-2.3`, update `master` branch versions from
+`2.3.0-SNAPSHOT` to `2.4.0-SNAPSHOT` in `pom.xml` files and update files like `R/pkg/DESCRIPTION`
+to refer to `2.4.0` rather than `2.3.0`.
```
HyukjinKwon: @srowen, I think we should change `__version__` in `python/pyspark/version.py`, say, `2.3.0.dev0` -> `2.4.0.dev0` (reminder - https://github.com/apache/spark/commit/9a145fd796145d1386fd75c01e4103deadb97ac9)

felixcheung: as note in the other PR, version changes are done automatically by `dev/create-release/release-tag.sh`?

(which is in the `Cutting a Release Candidate` section below)

basically when forking for a minor release, we want everything in release-tag to run in master except all the git commands (we don't want the tag in master but in release branch)

srowen: Sorry for the late reply. OK @felixcheung you're saying this is (or should be) done by the script? I'll just revert this part.

felixcheung:
Yes - there are multiple steps and I think this is best by refactoring release-tag into a new script that does the version update without making a git tag

PR #87, opened 2018-01-16
[SPARK-23083] [K8S] Add k8s as a supported option

foxish:
To be merged after 2.3 is released.

cc/ @rxin @srowen


site/sitemap.xml
```diff
@@ -672,27 +672,27 @@
   <changefreq>weekly</changefreq>
 </url>
 <url>
-  <loc>https://spark.apache.org/graphx/</loc>
+  <loc>https://spark.apache.org/sql/</loc>
```
srowen: @foxish could you revert this part? it won't matter much either way. I think it's because people are using slightly different versions of jekyll or something (update to the latest if you're not already). This file keeps changing back and forth otherwise.

srowen (approved):
Looks fine otherwise.

felixcheung (approved):

felixcheung:
btw, why is hbase in this image https://github.com/apache/spark-website/pull/87/files#diff-c53243cc9364367ffc84efa26ceb3bd2?

Also cassandra too I guess except for datastax-enterprise

srowen:
I think this is really "things Spark integrates with" rather than "runs on". I suppose it's not ambiguous -- it does the only thing it could do vis-a-vis HBase, Cassandra. But yeah I agree it's not obvious why that would be in the diagram.

felixcheung:
I guess these are Apache projects ... but then again there are so many more we could put on there.

site/sitemap.xml
foxish: Done

gatorsmile:
LGTM

srowen:
@foxish I merged this. You will have to close the PR manually in this case.

foxish:
Thanks!

PR #100, opened 2018-03-02
Redirect mean.html to column_aggregate_functions

shivaram:
This is a temporary fix to handle SparkR 2.3.0 where the vignette contains a link to mean.html but the page was removed in the release. 

https://github.com/apache/spark/pull/20711 fixes the problem for future releases but users who download the current SparkR 2.3.0 package will see a 404.


shivaram:
cc @felixcheung

felixcheung:
LGTM thanks!

PR #104, opened 2018-03-16
Add a project history page

mateiz:
This moves our history text to its own page and adds it in the nav structure, which people suggested to make it easier to find. I also fixed some stale text and links throughout.


rxin:
Thanks. I've merged this. Can you close the PR?

PR #105, opened 2018-03-18
add merge script

felixcheung:
Ported the merge script from spark main repo

### DIFF
--- a/doap.rdf
+++ b/doap.rdf
@@ -35,9 +35,6 @@
     <programming-language>Java</programming-language>
     <programming-language>Scala</programming-language>
     <programming-language>Python</programming-language>
-    <programming-language>R</programming-language>
-    <programming-language>SQL</programming-language>
-    <programming-language>JDBC</programming-language>
     <category rdf:resource="http://projects.apache.org/category/big-data" />
   </Project>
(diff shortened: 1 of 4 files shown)


doap.rdf
```diff
@@ -35,6 +35,9 @@
     <programming-language>Java</programming-language>
     <programming-language>Scala</programming-language>
     <programming-language>Python</programming-language>
+    <programming-language>R</programming-language>
```
srowen: This is not directly related right? but fine

srowen (approved):
Although duplication is a bit undesirable there's no easy way to share it otherwise. Seems fine, especially if you were able to trim out parts of the script that don't apply.

doap.rdf
HyukjinKwon: Hmmm .. I'd leave this change out though. Seems we need to fix 31L too.

HyukjinKwon (approved):

doap.rdf
felixcheung: haha ok. actually I'm not sure R is supported as a value, couldn't open some broken links

```
Not Found
The requested URL /languages.html was not found on this server.
```
https://projects.apache.org/languages.html

PR #106, opened 2018-03-28
Third-party-projects - mention Mist

dos65:


srowen (changes requested):
@dos65  It's fine, but please revert the change to `sitemap.xml`. I think people have different versions of the generator plugin installed or something.

dos65:
Done

PR #110, opened 2018-04-09
Add link to ApacheCon per Branding; fix a few other branding standards issues; fix sitemap.xml refs to localhost

srowen:


srowen:
Here's how the new events link looks at the moment: 
screen shot 2018-04-09 at 11 46 01

_layouts/global.html
```diff
@@ -211,10 +216,12 @@ <h5>Latest News</h5>
 
 <footer class="small">
   <hr>
-  Apache Spark, Spark, Apache, and the Spark logo are <a href="{{site.baseurl}}/trademarks.html">trademarks</a> of
-  <a href="https://www.apache.org">The Apache Software Foundation</a>.
+  Apache Spark, Spark, Apache, the Apache feather logo, and the Apache Spark project logo are either registered
```
srowen, line 219: This new text is pulled from the trademarks page and should make https://whimsy.apache.org/site/project/spark happier.

index.md
```diff
@@ -181,9 +181,3 @@ navigation:
     </ul>
   </div>
 </div>
-
-<div class="row">
```
srowen, line 185: This is a redundant download link at the bottom; not strictly related, but, I removed it as it's already in the right nav.

rxin:
LGTM

mateiz:
@srowen this is okay, but I would've put it below the Download Spark button since it will be a permanent fixture on the page. What do you think?

srowen:
My thinking was that it went more naturally with the block of news, which frequently announces events. If you have a moderate preference for the other way I can just change it.

PR #112, opened 2018-04-10
update doap file

felixcheung:


HyukjinKwon (approved):
Just wondering if we should wrote JDBC and SQL here in `description` too. LGTM anyway.

vanzin:
Merging.

PR #117, opened 2018-06-12
Update list of mirrored releases, add more instructions to release guide.

vanzin:


tgravescs:
changes look fine to me but I'm just starting to get familiar with release process

site/doap.rdf
```diff
@@ -28,13 +28,16 @@
...
+    <description>Apache Spark is a fast and general engine for large-scale data processing. It offers high-level APIs in Java, Scala, Python and R, as well as a rich set of libraries including stream processing, machine learning, and graph analytics.</description>
     <bug-database rdf:resource="http://issues.apache.org/jira/browse/SPARK" />
     <mailing-list rdf:resource="http://spark.apache.org/community.html" />
     <download-page rdf:resource="http://spark.apache.org/downloads.html" />
     <programming-language>Java</programming-language>
     <programming-language>Scala</programming-language>
     <programming-language>Python</programming-language>
+    <programming-language>R</programming-language>
+    <programming-language>SQL</programming-language>
+    <programming-language>JDBC</programming-language>
```
felixcheung, line 40: 👍

site/release-process.html
```diff
@@ -391,9 +391,17 @@ <h4>Remove Old Releases from Development Repository and Mirror Network</h4>
 <pre><code>svn rm https://dist.apache.org/repos/dist/dev/spark/v2.3.1-rc1-bin/ \
   https://dist.apache.org/repos/dist/dev/spark/v2.3.1-rc1-docs/ \
   -m"Removing RC artifacts."
-
 </code></pre>
 
+<p>Make sure to also remove the unpublished staging repositories from the <a href="https://repository.apache.org/">Apache Nexus Repository Manager</a>.</p>
+
+<p>When removing older releases, also update <code>js/download.js</code> to indicate the release is not mirrored
```
felixcheung: could we expand on `When removing older releases` happens?
I think given major.minor.patch, the previous patch release is archived when a newer one in the same major.minor is released?
ie, 2.1.3 released, 2.1.2 archived

felixcheung (approved):

site/release-process.html
vanzin: This is already sort of covered by the existing text:

    Spark always keeps two releases in the mirror network: the most recent release on the current and
    previous branches.

That doesn't seem to really correspond to the current mirrored releases (and with the expectation of the Hive tests that were added in 2.2), though, so maybe we should update that policy.

felixcheung: right, that doesn't seem to match the ASF mirror behavior now

felixcheung (approved):
is there removal on  https://dist.apache.org/repos/dist/release/spark/ automatic, by ASF mirror policy, or manual?
I see 2.3.0 now when 2.3.1 is up. maybe there's a time delay

vanzin:
It is manual. I did not remove 2.3.0 but maybe I should given the updated policy I just sort of made up.

felixcheung:
I see..

vanzin:
I'm merging this. We can continue to tweak the release guide as we find new things.

PR #119, opened 2018-06-28
[SPARK-24679] Download page should not link to unreleased code

lresende:
Remove instructions to checkout code which will download the
full repository including unreleased code. Also remove links
to nightly builds as these are not released code.

### DIFF
--- a/downloads.md
+++ b/downloads.md
@@ -53,17 +53,6 @@ Spark artifacts are [hosted in Maven Central](https://search.maven.org/#search%7
 ### Installing with PyPi
 <a href="https://pypi.python.org/pypi/pyspark">PySpark</a> is now available in pypi. To install just run `pip install pyspark`.
 
-### Spark Source Code Management
-If you are interested in working with the newest under-development code or contributing to Apache Spark development, you can also check out the master branch from Git:
-
-    # Master development branch
-    git clone git://github.com/apache/spark.git
-
-    # Maintenance branch with stability fixes on top of Spark 2.3.1
-    git clone git://github.com/apache/spark.git -b branch-2.3
-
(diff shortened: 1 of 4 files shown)


downloads.md
```diff
@@ -53,17 +53,6 @@ Spark artifacts are [hosted in Maven Central](https://search.maven.org/#search%7
 ### Installing with PyPi
 <a href="https://pypi.python.org/pypi/pyspark">PySpark</a> is now available in pypi. To install just run `pip install pyspark`.
 
-### Spark Source Code Management
```
vanzin, line 56: Could you move this or some version of it to `contributing.md` instead? That page doesn't seem to actually cover this.

lresende: Thanks @vanzin for the feedback, I have moved the cloning instructions to `contributing.md` and the info about nightly builds is already in `developer-tools.md`

vanzin:
Thanks, merging.

downloads.md
```diff
@@ -72,6 +61,4 @@ Once you've downloaded Spark, you can find instructions for installing and build
 
 As new Spark releases come out for each development stream, previous ones will be archived, but they are still available at [Spark release archives](https://archive.apache.org/dist/spark/).
 
-### Nightly Packages and Artifacts
```
srowen, line 75: Very belated, but, we had discussed this a while ago and agreed with other ASF folks that this link was OK.  It seems fine to move it to 'contributing' (see http://www.apache.org/legal/release-policy.html#host-rc ) but seems like it was just deleted.

lresende: As mentioned above, the nightly builds is already available at [developer tools page](http://spark.apache.org/developer-tools.html), that's why it was just removed from here. Are you suggesting we add it to contributing as well ?

srowen: Ah sorry I did miss that, yes this fine.

PR #120, opened 2018-06-29
Update website for 2.1.3 release.

vanzin:


vanzin:
3 commits:
- generated docs from 2.1.3 release
- updated website source
- generated web site content

@holdenk could you take care of updating this if needed, and merging it when the mirrors are updated? I won't be around to look at it. Thanks!

HyukjinKwon:
@vanzin, just for doubly sure, did you generate it with Python 3's Sphinx? There was an issue found [SPARK-24530](https://issues.apache.org/jira/browse/SPARK-24530) and I was trying to update the other documentation, release process here, and probably the Makefile script to add a control of Python executable version.

holdenk:
@HyukjinKwon checked the docs, they exhibit the same *args/ **kwargs behaviour that SPARK-24530 describes as a bug, however it's not a regression and we already voted on these docs. We could rebuild the docs at a later point if we all decide that's worth the risk (e.g. possible broken links if we aren't careful).

holdenk (approved):
LGTM ran locally with Jekyll serve.

HyukjinKwon:
It's actually a regression ... will update the doc soon anyway.

holdenk:
It's only a regression if it worked in the previous version, as the links in the JIRA indicate this was not the case since it did not work in `2.1.2`. If you update the docs please be very careful not to break any links.

HyukjinKwon:
It's still a step back comparing 2.1.1 which is found later now (for clarification, I am not arguing we should have blocked the release). Yup, will do.

PR #122, opened 2018-07-02
Update release process to use Python 3 for Python API documentation

HyukjinKwon:
This PR proposes to set `SPHINXPYTHON` environment variable to work around missing signatures.

Note that this PR is dependent on https://github.com/apache/spark/pull/21659.


release-process.md
```diff
@@ -86,6 +86,7 @@ Instead much of the same release logic can be accessed in `dev/create-release/re
 - R, for CRAN packaging tests, requires e1071 to be installed as part of the packaging tests.
 - In addition R uses LaTeX for some things, and requires some additional fonts. On Debian based systems you may wish to install `texlive-fonts-recommended` and `texlive-fonts-extra`.
 - Make sure you required Python packages for packaging (see `dev/requirements.txt`)
```
felixcheung, line 88: should we put that (sphinx) into dev/requirements.txt?

HyukjinKwon: Indeed yea I think we should.

PR #125, opened 2018-07-02
Fix signature description broken in PySpark API documentation in 2.3.1

HyukjinKwon:
This PR proposes to fix signature description broken in PySpark API documentation in 2.3.1. See SPARK-24530.

Sphinx version was intentionally matched to reduce the diff.


mengxr:
Could you split this PR into two, one fixing the version and the other fixing the ML doc?

PR #135, opened 2018-08-12
Stash pride logo for next year

srowen:
OK to return to the normal logo until next year? to keep it special.


srowen:
CC @holdenk

tgravescs:
+1. Makes sense to me.

PR #136, opened 2018-08-14
Mention correctness issues as blockers in developer docs; mention 18 month release branch maintenance guideline

srowen:


srowen:
CC @tgravescs @squito

contributing.md
```diff
@@ -258,6 +258,10 @@ Example: `Fix typos in Foo scaladoc`
         1. **Component**
         1. **Affects Version**. For Bugs, assign at least one version that is known to exhibit the 
         problem or need the change
+        1. **Label**. Not widely used, except for the following:
+             - `correctness`: a data loss or correctness issue
```
felixcheung: I feel perhaps we should use a different label for data loss?

contributing.md
```diff
@@ -258,6 +258,10 @@ Example: `Fix typos in Foo scaladoc`
...
+             - `release-notes`: the change's effects need mention in release notes
```
tgravescs: For this one do we have an official way to write the text that would go into the release notes?

srowen: Probably a list of JIRAs and short summary of the impact of each so that readers can find things that will affect them and read details. See how behavior changes are handled here: https://spark.apache.org/releases/spark-release-2-3-0.html

versioning-policy.md
```diff
@@ -54,7 +54,7 @@ available APIs.
 
 In general, feature ("minor") releases occur about every 6 months. Hence, Spark 2.3.0 would
 generally be released about 6 months after 2.2.0. Maintenance releases happen as needed.
-A minor release usually sees 1-2 maintenance releases in the 6 months following its first
+A feature release usually sees a maintenance release in the 6 months following its first
```
tgravescs: not sure "sees a" makes sense here, we usually do multiple

Maybe we can just get rid of this sentence or reword more like "Maintenance releases happen as needed in between the feature releases.

versioning-policy.md
```diff
@@ -64,3 +64,9 @@ release. Major releases do not happen according to a fixed schedule.
 | Mid Aug 2018 || Code freeze. Release branch cut.|
 | Late Aug 2018 || QA period. Focus on bug fixes, tests, stability and docs. Generally, no new features merged.|
 | Early Sep 2018 || Release candidates (RC), voting, etc. until final release passes|
+
+<h2>Maintenance Releases and EOL</h2>
+
+Feature release branches will, generally, be maintained with bug fixes releases for a period of 18 months.
```
tgravescs: "bug fix releases"

contributing.md
gatorsmile: Data loss is part of correctness issues, right?

contributing.md
gatorsmile: All the JIRAs that change the behavior needs to be mentioned in release notes. Thus, it is marked as release-notes too. We started this since 2.3.0.

versioning-policy.md
```diff
@@ -64,3 +64,9 @@ release. Major releases do not happen according to a fixed schedule.
 | Mid Aug 2018 || Code freeze. Release branch cut.|
 | Late Aug 2018 || QA period. Focus on bug fixes, tests, stability and docs. Generally, no new features merged.|
 | Early Sep 2018 || Release candidates (RC), voting, etc. until final release passes|
+
+<h2>Maintenance Releases and EOL</h2>
+
+Feature release branches will, generally, be maintained with bug fixes releases for a period of 18 months. 
+For example, branch 2.1.x is no longer considered maintained as of July 2018, 18 months after the release 
+of 2.1.0 in December 2016. No more 2.1.x releases should be expected after that point, even for bug fixes.
```
tgravescs: I'm fine with the 18 months, we should make sure everyone agrees. This essentially means we will be keeping 3 releases active.  And at this point 2.1.x will be dead. Did this go to the dev list?  I apologize if I missed it seems my yahoo mail account hasn't been getting everything for some reason.

contributing.md
vanzin: I think we need either the PR author or the person who committed the change to actually write some release note blurb for the RM to include. It's not trivial, at times, for the RM to go and figure out what should be in the release note.

(Internally in our jira we have a special field for that, not sure whether we can get that in Apache's instance.)

tgravescs: I agree with I think we need to make sure the jira title is clear and then have somewhere we put more specific text about the change so a user can easily find it.  If Apache won't let us do this perhaps in the description of the jira we can add our own section that is like ========== RELEASE NOTES ============

Is this only for "Changes of behavior" then?

contributing.md
srowen: Let's suggest both as viable labels. We'd obviously pay attention to both, and if a data-loss issue is labeled correctness, oh well. Might give people a little more descriptiveness at no real additional complexity.

contributing.md
gatorsmile: In Spark SQL, we require the PR author to write the behavior changes in the same PR. https://github.com/apache/spark/blame/master/docs/sql-programming-guide.md#L1877

It is very easy for RM to write the release note. How about the other components? I think we can follow the same rule?

vanzin: I don't think that's a good approach. Sometimes the release note label is added after the bug is fixed, and the PR has no information about what should be in the release note (e.g. SPARK-24771).

Sometimes the commit message just says "something should be mentioned in the release notes", instead of actually having something that can be copy & pasted into the release notes. I ran into something like that for 2.3.1 (SPARK-23173). It's much better if the people who are familiar with the change write the actual release note, instead of asking the RM to become familiar with code they may have never seen before and hope they understood what's going on.

It shouldn't be hard to add the note to jira itself. Even if just means you copy & paste a little bit of text from the commit message.

srowen (commented):
I will send a note to dev@ about the 18-month point. I did mention it as part of the thread you started but needs more highlighting.

contributing.md
gatorsmile: We need to record the behavior changes in both the doc and the release notes. It is easier for end users to check the behavior changes. That will be a Single Source of Truth; otherwise, they have to check all the release notes, if they migrate from 2.0 to 2.3. If the release note is out, it is hard for the other to edit the release note. However, we still can change the doc easily. This is what Spark SQL is doing.

vanzin: I'm not sure I follow what you're saying.

It is fine if you want to keep some not in the PR description. What I'm saying is that:

- we should have release notes for the end user in the release announcements we have on the site
- that we should reduce the burden of the RM to generate those release notes.

The current approach sucks for the RM, because they have to go and figure out exactly what to write down in those documents. What I'm proposing is that the people involved in the change actually create that entry, and the RM's job becomes just to collect them when creating the release announcement.

The bug is a perfect place for that, because it can be updated after the change is committed, something that is not true for the commit message. So, if you want, by all means, keep asking the PR author to write something in the PR. But that shouldn't be the source of truth; just add that as a release note in jira, and it can be edited as needed, and it's much easier for the RM to find.

contributing.md
```diff
@@ -247,7 +247,7 @@ Example: `Fix typos in Foo scaladoc`
         committers to set. JIRA tends to unfortunately conflate "size" and "importance" in its 
         Priority field values. Their meaning is roughly:
              1. Blocker: pointless to release without this change as the release would be unusable 
-             to a large minority of users
+             to a large minority of users. Correctness and data loss issues should be considered Blockers.
```
dongjoon-hyun: Since this is `contributing.md`, the process looks like the followings.
1. Contributor can only make a JIRA issue with `Major` with `correctness` label.
2. Committers should validate this and promote it from `Major` to `Blocker`.

It's a gray area during the period between (1) and (2). Or, are we widely open for contributors to create `Blocker` issue with `correctness` label?

contributing.md
srowen: (This is what I am saying too, or trying to, in the doc. All, feel free to propose better wording.)

contributing.md
srowen: I don't mind anyone setting Blocker if they really think it's a correctness or data-loss issue. I'll clarify that.

contributing.md
vanzin: I think what's missing here is how to provide the information about what's changing. The label just tells us that something is changing, but right now whoever is looking for the release notes (the RM when preparing the release, or some user digging for this stuff) doesn't know where to look for it.

What I'm proposing is that we have a field in jira (or some specially marked comment like Tom suggested) where the description of the changing behavior resides. Whether that's written after the fact or copy & pasted from the commit message, it doesn't matter. I just strongly wish that this information was in the bug itself, where it's much easier to find when coupled with the label.

srowen: Agree, don't know how to get a new field added; INFRA could help but don't know how easy it is. We can, at least, just say here that this info needs to be in the JIRA or PR?

vanzin: I think it must be in jira, and it can be in the PR (in which case the person closing the bug can copy & paste it).

vanzin: I just checked jira and there's a "Docs Text" field we could re-purpose for this, since it doesn't look like we use it for anything currently.

vanzin:
LGTM. Perhaps clarify that "Docs Text" can (should?) be set when the bug is marked as fixed.

contributing.md
gatorsmile: Regarding the behavior changes in Spark SQL, the job of RM is simple. They just need to copy and paste what we added in the sql doc.

vanzin: We shouldn't have different policies for different modules. A release note for core is as important as one for SQL. And right now we don't have a defined policy for anything non-SQL.

If you want to make the SQL approach apply to everything else, then great. Propose that and change the release manager guide, which doesn't mention the SQL stuff at all.

PR #137, opened 2018-08-14
For projects using names that are likely proscribed, either update to use current conforming name or remove the link, for now

srowen:


site/third-party-projects.html
```diff
@@ -211,7 +217,7 @@ <h2>spark-packages.org</h2>
 <h2>Infrastructure Projects</h2>
 
 <ul>
-  <li><a href="https://github.com/spark-jobserver/spark-jobserver">Spark Job Server</a> - 
+  <li><a href="https://github.com/spark-jobserver/spark-jobserver">REST job server for Apache Spark</a> -
```
mateiz: Nit: maybe capitalize "Job Server"

site/third-party-projects.html
```diff
@@ -220,25 +226,17 @@ <h2>Infrastructure Projects</h2>
...
-  <li><a href="https://github.com/datastax/spark-cassandra-connector">Spark Cassandra Connector</a> - 
-Easily load your Cassandra data into Spark and Spark SQL; from Datastax</li>
   <li><a href="https://github.com/filodb/FiloDB">FiloDB</a> - a Spark integrated analytical/columnar 
 database, with in-memory option capable of sub-second concurrent queries</li>
   <li><a href="http://www.elasticsearch.org/guide/en/elasticsearch/hadoop/master/spark.html#spark-sql">ElasticSearch - 
 Spark SQL</a> Integration</li>
-  <li><a href="https://github.com/tresata/spark-scalding">Spark-Scalding</a> - Easily transition 
-Cascading/Scalding code to Spark</li>
   <li><a href="http://zeppelin-project.org/">Zeppelin</a> - an IPython-like notebook for Spark. There 
 is also <a href="https://github.com/tribbloid/ISpark">ISpark</a>, and the
```
mateiz: ISpark seems unmaintained so I'd remove it.

mateiz: Spark Notebook is also problematic w.r.t the trademark policy, so we should reach out to it as well.

mateiz:
Looks good to me modulo some comments on the notebook ones, but let's see whether other people have comments.

third-party-projects.md
```diff
@@ -9,6 +9,12 @@ navigation:
 
 This page tracks external software projects that supplement Apache Spark and add to its ecosystem.
 
+To add a project, open a pull request against the [spark-website](https://github.com/apache/spark-website) 
+repository. Add an entry to this markdown file, then run `jekyll build` to generate the HTML too. Include
```
felixcheung: maybe point to the markdown file directly here instead of "this markdown file"?

third-party-projects.md
```diff
@@ -72,7 +68,7 @@ transforming, and analyzing genomic data using Apache Spark
 
 <h3>C# / .NET</h3>
 
-- <a href="https://github.com/Microsoft/SparkCLR">CLR for Spark</a>
+- <a href="https://github.com/Microsoft/Mobius">Mobius</a>: C# and F# language binding and extensions to Apache Spark
```
felixcheung, line 72: tbh this isn't being maintained either... (the last commit is Apr 2017)

felixcheung (commented):
I'm not sure if I'm fully understand the guideline on this.
for example is `clj-spark` ok? is `Sparkling` ok? is `Spark.jl` ok?

site/third-party-projects.html
felixcheung: I'd not call Zeppelin `an IPython-like notebook for Spark` either.

maybe `Multi-purpose notebook which supports 20+ language backends, including Apache Spark`  https://zeppelin.apache.org/docs/0.8.0/

third-party-projects.md
srowen: Will make the changes. Things like spark.jl are ambiguous because they are software identifiers (OK) that become de facto project names (not OK). 'Sparkling' is something the guidelines say is proscribed but I honestly think that's more of a gray area, for the ASF to claim the trademark extends to a different English word. We can remove anything that is in doubt here.

third-party-projects.md
srowen: Meh, that could possibly be considered not-obsolete. If it's not a problematic name, and not clearly dead, I figure we can keep it for now?

third-party-projects.md
srowen: Hm, although Spark's trademark policy says things like "sparkly" aren't OK, I don't really see that in https://www.apache.org/foundation/marks/ ? It doesn't strike me as confusingly similar.

mateiz: We've received the feedback that these aren't OK on some previous trademarks@ threads, but it could be good to clarify. It's true that the policy doesn't seem to list examples, at least on that page.

third-party-projects.md
felixcheung: ok

felixcheung (approved):

PR #148, opened 2018-09-19
Added the Spark Operator as a third-party project

liyinan926:
@srowen.

### DIFF
--- a/site/third-party-projects.html
+++ b/site/third-party-projects.html
@@ -239,6 +239,7 @@ against Spark, and data scientists to use Javascript in Jupyter notebooks.</li>
   <li><a href="https://github.com/SnappyDataInc/snappydata">SnappyData</a> - an open source 
 OLTP + OLAP database integrated with Spark on the same JVMs.</li>
   <li><a href="https://github.com/Hydrospheredata/mist">Mist</a> - Serverless proxy for Spark cluster (spark middleware)</li>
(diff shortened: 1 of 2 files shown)


srowen:
@liyinan926 actually I believe the PMC would say that project names like "Spark Operator" aren't allowed under ASF trademark guidelines. Could the description be changed to rename it "K8S Operator for Apache Spark" or similar?

liyinan926:
@srowen the description has been fixed.

srowen:
It looks like you still need to regenerate the HTML. The project site still says Spark Operator too. The project name would need to be something like 'Operator for Apache Spark'

liyinan926:
Oops, forgot to rebuild. Fixed now.

srowen:
I am talking about https://github.com/GoogleCloudPlatform/spark-on-k8s-operator , and still don't see updates there.

liyinan926:
You meant the name of the repo `spark-on-k8s-operator` or the appearances of the term `Spark Operator` in project description and documentation?

srowen:
See https://spark.apache.org/trademarks.html Generally there's an exception for software identifiers like a repo or artifact or package name string. But yes I am otherwise talking about how the project is titled. 'Spark Operator' would not be consistent with these guidelines.

liyinan926:
Thanks for the info. Will rename it to `K8S Operator for Apache Spark` in the documentation. So the repo name is fine, right?

srowen:
Yes, although ideally projects just choose a name without Spark at all and therefore don't need it in package or repo IDs, for those that do, it's recognized that using a long ID with '-for-apache-spark' is cumbersome.

liyinan926:
The project documentation has been fixed to use `Kubernetes Operator for Apache Spark` instead of `Spark Operator`.

srowen (approved):

PR #151, opened 2018-09-29
Add a section to update test suites

dongjoon-hyun:


dongjoon-hyun:
cc @HyukjinKwon , @gatorsmile , @srowen

site/release-process.html
```diff
@@ -511,6 +511,20 @@ <h4>Create an Announcement</h4>
...
 
+<h4>Update a test suite depending on old releases</h4>
+
+<p>Since Spark always keeps the latest maintance released of each branch in the mirror network,
+we need to update a test suite to check backward compatibility with downloading old versions.
+For now, there is one test suite, <code>HiveExternalCatalogVersionsSuite</code>.</p>
+
+<p>For a new maintenance release, <code>HiveExternalCatalogVersionsSuite</code> will not fail.
+However, it will cause many trials and fallbacks in all Jenkins builds over corresponding branches.
+This updates should be done as a separate PR after release announcement. (See SPARK-25570)</p>
```
HyukjinKwon: `updates` -> `update`

HyukjinKwon (approved):

release-process.md
```diff
@@ -319,4 +319,18 @@ Once everything is working (website docs, website changes) create an announcemen
 and then send an e-mail to the mailing list. To create an announcement, create a post under
 `news/_posts` and then run `jekyll build`.
 
+<h4>Update a test suite depending on old releases</h4>
+
+Since Spark always keeps the latest maintance released of each branch in the mirror network,
```
srowen: maintance released -> maintenance release
and
with downloading -> by testing against (?)

release-process.md
```diff
@@ -319,4 +319,18 @@ Once everything is working (website docs, website changes) create an announcemen
...
+we need to update a test suite to check backward compatibility with downloading old versions.
+For now, there is one test suite, `HiveExternalCatalogVersionsSuite`.
+
+For a new maintenance release, `HiveExternalCatalogVersionsSuite` will not fail.
+However, it will cause many trials and fallbacks in all Jenkins builds over corresponding branches.
+This updates should be done as a separate PR after release announcement. (See SPARK-25570)
```
srowen: I understand what this text is about, but I'm not sure others would get it from this description. Basically, this test suite should refer to only the most recent release from each actively maintained branch, right? I think it's worth giving a short example of what this means.

gatorsmile: I think the above description is not needed. Just explain what the release managers need to do. This is Spark SQL specific. I am afraid this might confuse the release managers who are not familiar with Spark SQL.

release-process.md
dongjoon-hyun: Thanks!

dongjoon-hyun:
Thank you, @HyukjinKwon , @srowen , @gatorsmile . I updated like the followings.
- `updates` -> `update`
- `maintance released` -> `maintenance release`
- `with downloading` -> `by testing against`
- Remove `This updates should be done as a separate PR after release announcement. (See SPARK-25570)`

gatorsmile:
Generally, it looks fine to me. cc @cloud-fan who is the original creator of this test suite.

gatorsmile:
BTW, the next three days are the national holiday of @cloud-fan . He might not reply the ping very quickly.

release-process.md
```diff
@@ -319,4 +319,17 @@ Once everything is working (website docs, website changes) create an announcemen
 and then send an e-mail to the mailing list. To create an announcement, create a post under
 `news/_posts` and then run `jekyll build`.
 
+<h4>Update a test suite depending on old releases</h4>
```
srowen: I think we can say "HiveExternalCatalogVersionsSuite" here instead of "a test suite" to be more direct. If it's many test suites later, we can change this.

release-process.md
```diff
@@ -319,4 +319,17 @@ Once everything is working (website docs, website changes) create an announcemen
...
+
+Since Spark always keeps the latest maintenance release of each branch in the mirror network,
```
srowen: Maybe "the Spark project maintains _only_ the _latest_ maintenance release ..."?

release-process.md
```diff
@@ -319,4 +319,17 @@ Once everything is working (website docs, website changes) create an announcemen
...
+we need to update a test suite to check backward compatibility by testing against old versions.
```
srowen: Just say `HiveExternalCatalogVersionsSuite` here too I think and remove the next sentence.

release-process.md
```diff
@@ -319,4 +319,17 @@ Once everything is working (website docs, website changes) create an announcemen
...
+For now, there is one test suite, `HiveExternalCatalogVersionsSuite`.
+
+For a new maintenance release, `HiveExternalCatalogVersionsSuite` will not fail.
+However, it will cause many trials and fallbacks in all Jenkins builds over corresponding branches.
+
+For a new minor release, we need to add the new version to preserve the test coverage. (See SPARK-23570)
```
srowen: This points to the right change, but, how about just saying what has to change directly? It's easy to describe.

When a new release occurs, `PROCESS_TABLES.testingVersions` in `HiveExternalCatalogVersionsSuite` must be updated shortly thereafter. This list should contain the latest release in all active maintenance branches, and no more. For example, as of this writing, it has value `val testingVersions = Seq("2.1.3", "2.2.2", "2.3.2")`. "2.4.0" will be added to the list when it's released. "2.1.3" will be removed (and removed from the Spark dist mirrors) when the branch is no longer maintained. "2.3.2" will become "2.3.3" when 2.3.3 is released.

srowen (changes requested):

release-process.md
dongjoon-hyun: Yep.

release-process.md
dongjoon-hyun: Thanks! I'll update like that.

dongjoon-hyun:
The section is simplified by describing directly as @srowen advised. Also, it's moved before the last section `Create an Announcement`.

srowen (approved):

cloud-fan:
LGTM

dongjoon-hyun:
Thank you all. I'll merge this.

PR #153, opened 2018-10-05
improve the document about contributing to bug reports

cloud-fan:
Recently we hit many correctness bugs, and some of them do not get enough attention. I want to highlight the importance of data correctness/data loss issues, and suggest contributors to describe how the bug was introduced, so that people can quickly understand the bug. It's also possible that the reporter doesn't realize it's a correctness bug, by looking at how the bug was introduced, other people can point it out.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -89,7 +89,8 @@ first. Unreproducible bugs, or simple error reports, may be closed.
 
 It's very helpful if the bug report has a description about how the bug was introduced, by 
 which commit, so that reviewers can easily understand the bug. It also helps committers to 
-decide how far the bug fix should be backported, when the pull request is merged.
+decide how far the bug fix should be backported, when the pull request is merged. The pull 
+request to fix the bug should narrow down the problem to the root cause.
 
 Performance regression is also one kind of bug. The pull request to fix a performance regression 
 must provide a benchmark to prove the problem is indeed fixed.
--- a/contributing.md
+++ b/contributing.md
@@ -87,6 +87,16 @@ and ideally reproduce the bug. Simply encountering an error does not mean a bug 
 reported; as below, search JIRA and search and inquire on the Spark user / dev mailing lists 
 first. Unreproducible bugs, or simple error reports, may be closed.
 
+It's very helpful if the bug report has a description about how the bug was introduced, by 
+which commit, so that reviewers can easily understand the bug. It also helps committers to 
+decide how far the bug fix should be backported, when the pull request is merged.
+
+Performance regression is also one kind of bug. The pull request to fix a performance regression 
+must provide a benchmark to prove the problem is indeed fixed.
+
+Note that, data correctness/data loss bugs are very serious. If the bug report doesn't get 
+enough attention, please send an email to `dev@spark.apache.org`, to draw more attentions.
+
 It is possible to propose new features as well. These are generally not helpful unless 
 accompanied by detail, such as a design document and/or code change. Large new contributions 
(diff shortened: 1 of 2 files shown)


cloud-fan:
cc @rxin @srowen @tgravescs @gatorsmile

contributing.md
```diff
@@ -87,6 +87,16 @@ and ideally reproduce the bug. Simply encountering an error does not mean a bug
...
 
+It's very helpful if the bug report has a description about how the bug was introduced, by 
+which commit, so that reviewers can easily understand the bug. It also helps committers to 
+decide how far the bug fix should be backported, when the pull request is merged.
+
+Performance regression is also one kind of bug. The pull request to fix a performance regression 
+must provide a benchmark to prove the problem is indeed fixed.
+
+Note that, data correctness/data loss bugs are very serious. If the bug report doesn't get 
+enough attention, please send an email to `dev@spark.apache.org`, to draw more attentions.
```
HyukjinKwon, line 100: @cloud-fan, how about we document tagging things for both tags together while we are here so that contributors mark the tags correspondingly?

contributing.md
```diff
@@ -87,6 +87,16 @@ and ideally reproduce the bug. Simply encountering an error does not mean a bug
 reported; as below, search JIRA and search and inquire on the Spark user / dev mailing lists 
 first. Unreproducible bugs, or simple error reports, may be closed.
 
+It's very helpful if the bug report has a description about how the bug was introduced, by 
+which commit, so that reviewers can easily understand the bug. It also helps committers to
```
HyukjinKwon, line 91: I would add a self-reproducer too .. or at least narrowing down the issue.

srowen (commented):
Seems like a fine kind of thing to add to the guidance.

contributing.md
```diff
@@ -87,6 +87,16 @@ and ideally reproduce the bug. Simply encountering an error does not mean a bug
 reported; as below, search JIRA and search and inquire on the Spark user / dev mailing lists 
 first. Unreproducible bugs, or simple error reports, may be closed.
 
+It's very helpful if the bug report has a description about how the bug was introduced, by 
+which commit, so that reviewers can easily understand the bug. It also helps committers to 
+decide how far the bug fix should be backported, when the pull request is merged.
+
+Performance regression is also one kind of bug. The pull request to fix a performance regression 
+must provide a benchmark to prove the problem is indeed fixed.
```
HyukjinKwon, line 96: It should provide a benchmark of course usually. But is it a must to comply? I would rather make it `should` and leave a room for exceptions.

contributing.md
cloud-fan: It's already mentioned in the paragraph above
> Bug reports are only useful however if they include enough information to understand, isolate 
 and ideally reproduce the bug.

contributing.md
cloud-fan: It's like a bug fix must provide a regression test. Without a benchmark, how can we know the performance regression is real and it's fixed? The benchmark doesn't need to be committed to spark codebase though.

HyukjinKwon: Ah, yea. I meant the commit stuff. The performance effect should be clarified in any form if a PR targets to fix a performance, of course.

contributing.md
HyukjinKwon: How about additionally mentioning narrowing down the problem? At least I spend some time to ask the information and narrow down the issue time to time.

HyukjinKwon:
+1 i like this

HyukjinKwon (approved):

PR #157, opened 2018-11-02
update the release-process document

cloud-fan:
Our document of doing release is out-dated. This PR updates the release doc to match the process of Spark 2.4.0 release.
1. cutting RC has been fully automated. People should always use the docker script, instead of setting up the environment manually. Then we don't need to update document everytime an environment change is needed.
2. when finalizing the release, suggest people to retain the generated docs of latest RC,  and copy it to spark-website, instead of re-generating it.


release-process.md
```diff
@@ -178,38 +169,6 @@ Publishing to CRAN is done using <a href="https://cran.r-project.org/submit.html
 Since it requires further manual steps, please also contact the <a href="mailto:private@spark.apache.org">PMC</a>.
 
 
-<h4> Remove RC artifacts from repositories</h4>
```
cloud-fan: I moved the cleanup process to the end, since the docs are still needed when updating spark-website

cloud-fan:
cc @vanzin @srowen @holdenk @felixcheung @jerryshao @gatorsmile

srowen (approved):

release-process.md
```diff
@@ -223,23 +182,16 @@ $ git push apache v1.1.1
...
 
 ```
-# Build the latest docs
-$ git checkout v1.1.1
-$ cd docs
-$ PRODUCTION=1 jekyll build
-
-# Copy the new documentation to Apache
+# copy the docs of the voted RC to spark-website
+$ svn co "https://dist.apache.org/repos/dist/dev/spark/v2.4.0-rc5-docs" spark-docs
```
vanzin: Since you may have this info locally, how many files are in the current docs build?

Last time I tried to checkout a huge directory from svn, the ASF infra guys blocked our network from talking to the ASF servers. So I'd avoid this if that directory is kinda large.

I don't remember if I actually did that or just thought about it, but one idea was to leave the docs build locally after the rc is generated, instead of cleaning it up when the docker script is gone. That would allow you to just upload your local copy to spark-website.

cloud-fan: For 2.4.0 it's 3572 files, 145 MB.

During a release, we upload these doc files many times(for each RC), and download them once. Will this be a problem?

cloud-fan: BTW during RC voting, if one people want to validate the doc, he has to check it out locally and start a server to look at the web pages. If the ASF infra blocks it, it will be hard for people to evaluate an RC.

site/release-process.html
```diff
@@ -505,6 +470,34 @@ <h4>Update the Spark Website</h4>
 $ git log v1.1.1 --grep "$expr" --shortstat --oneline | grep -B 1 -e "[3-9][0-9][0-9] insert" -e "[1-9][1-9][1-9][1-9] insert" | grep SPARK &gt; large-patches.txt
 </code></pre>
 
+<h4> Remove RC artifacts from repositories</h4>
+
+<p>After the vote passes and you moved the approved RC to the release repository, you should delete
+the RC directories from the staging repository. For example:</p>
+
+<pre><code>svn rm https://dist.apache.org/repos/dist/dev/spark/v2.3.1-rc1-bin/ \
```
felixcheung: the version string in the same should be consistent through out I think? shouldn't this be v2.4.0-rc5?

release-process.md
vanzin: That's a lot of files. I think the thing the ASF doesn't like is a burst of requests, and this might trigger it. Uploading is not a problem, since it's authenticated. Downloading is not, so it could be a DoS, which is why they do the blocking.

> if one people want to validate the doc, he has to check it out locally 

Actually they can see it directly on the ASF server without downloading it locally. Using a browser and manually looking at the logs automatically "throttles" the number of requests, so it shouldn't trigger any warnings to the infra guys.

I think it's safer to recommend keeping the docs for the last RC locally in case it passes, and only checkout from svn if necessary.

release-process.md
```diff
@@ -35,6 +35,34 @@ If you are a new Release Manager, you can read up on the process from the follow
...
+<a href="https://www.apache.org/dev/openpgp.html#export-public-key">https://www.apache.org/dev/openpgp.html#export-public-key</a>
+
+```
+# Move dev/ to release/ when the voting is completed. See Finalize the Release below
+svn co --depth=files "https://dist.apache.org/repos/dist/dev/spark" svn-spark
+# edit svn-spark/KEYS file
+svn ci --username $ASF_USERNAME --password "$ASF_PASSWORD" -m"Update KEYS"
+```
+
+<h3>Installing docker</h3>
```
jerryshao, line 59: Maybe we should add an another paragraph to do the release without docker. It is pretty similar to docker, just running the script `do-release.sh` is enough. The only difference is that user should manually install all the dependencies as listed in docker file.

cloud-fan: The thing I try to avoid is, maintain a document about how to set up the environment/dependencies. We should control it fully through the scripts.

I think we should not recommend people to do release without docker. If someone does want to run `do-release.sh` directly, he should read the docker file and other related scripts.

jerryshao: OK, I see.

release-process.md
```diff
@@ -35,6 +35,34 @@ If you are a new Release Manager, you can read up on the process from the follow
 - gpg for signing https://www.apache.org/dev/openpgp.html
 - svn https://www.apache.org/dev/version-control.html#https-svn
 
+<h3>Preparing gpg key</h3>
+
+You can skip this section if you have been a release manager before.
```
jerryshao: "if you have been a release manager before" => "if you've already uploaded your key". No need to be a release manager before :).

release-process.md
```diff
@@ -283,11 +283,14 @@ which fetches potential replacements from Github and JIRA. For instance:
...
+$ export RELEASE_TAG=v1.1.1
+$ export PREVIOUS_RELEASE_TAG=v1.1.0
 # Generate initial contributors list, likely with warnings
 $ ./generate-contributors.py
-# Set JIRA_USERNAME, JIRA_PASSWORD, and GITHUB_API_TOKEN
-$ vim release-spark/dev/translate-contributors.py
+# set JIRA_USERNAME, JIRA_PASSWORD, and GITHUB_API_TOKEN
+$ export JIRA_USERNAME=blabla
+$ export JIRA_PASSWORD=blabla
+$ export GITHUB_API_TOKEN=blabla
```
jerryshao, line 293: Do we require this?

release-process.md
```diff
@@ -71,45 +99,15 @@ Also check that all build and test passes are green from the RISELab Jenkins: ht
...
 
 
-The process of cutting a release candidate has been partially automated via the RISELab Jenkins. There are
-Jenkins jobs that can tag a release candidate and create various packages based on that candidate.
-At present the Jenkins jobs *SHOULD NOT BE USED* as they use a legacy shared key for signing.
+To cut a release candidate, there are 4 steps:
+1. Create a git tag for the release candidate.
+1. Package the release binaries & sources, and upload them to a staging SVN repo.
+1. Create the release docs, and upload them to a staging SVN repo.
+1. Publish a snapshot to the Apache release repo.
```
jerryshao: Sorry what's the meaning of this step, I don't remember I have a such step.

felixcheung: `- Publish a snapshot to the Apache release repo `dev/create-release/release-build.sh publish-release``

felixcheung: I do think this is a bit confusing for new RM. Perhaps `Apache release repo` should be `Apache staging Maven repo`

jerryshao: Ohh, you mean pushing artifacts to apache staging maven repo? This seems confusing.

cloud-fan: This is how we describe it before: https://github.com/apache/spark-website/pull/157/files#diff-12257e3523e2ce22fccd80687597e7cdL94

+1 that "Apache staging Maven repo" is cleaner, I'll update soon

release-process.md
cloud-fan: I think so. It needs to access github data to get a people's full name.

jerryshao: I see. Got it.

vanzin:
LGTM

HyukjinKwon (approved):

holdenk:
Thank for updating the release process docs!

cloud-fan:
I'm having a problem to merge this PR. The merge script needs to know the original apache git repo, but do we have one for spark-website?

srowen:
Setup should be the same as for the spark repo @cloud-fan ; here's mine for example:

```
$ git remote -v
apache	https://git-wip-us.apache.org/repos/asf/spark-website.git (fetch)
apache	https://git-wip-us.apache.org/repos/asf/spark-website.git (push)
apache-github	git://github.com/apache/spark-website (fetch)
apache-github	git://github.com/apache/spark-website (push)
origin	https://github.com/srowen/spark-website.git (fetch)
origin	https://github.com/srowen/spark-website.git (push)
upstream	https://github.com/apache/spark-website.git (fetch)
upstream	https://github.com/apache/spark-website.git (push)
```

cloud-fan:
thanks @srowen , it works!

felixcheung:
FYI, there is a merge script: https://github.com/apache/spark-website/blob/asf-site/merge_pr.py
EDIT: sorry, I might have misunderstood. is it the remote repo needs to be added?

cloud-fan:
yes we need to add the remote repo, which I didn't know before

PR #158, opened 2018-11-05
Apache Spark 2.4.0 Release: Announcements & Release Notes

cloud-fan:
This patch adds the release notes and relevant announcements for Spark 2.4.0.

The first commit just adds the docs of the latest RC, please review the second commit.

Release Notes: https://github.com/cloud-fan/spark-website/blob/62847a7d4b106ad6ac1aee2016e226b56c7e9d1a/releases/_posts/2018-11-05-spark-release-2-4-0.md


cloud-fan:
cc @rxin @vanzin @srowen @holdenk @felixcheung @gatorsmile

felixcheung (approved):
it's too many to list or review on github - I reviewed those not under site/docs, LGTM

releases/_posts/2018-11-05-spark-release-2-4-0.md
```diff
@@ -0,0 +1,188 @@
...
+* This will become a table of contents (this text will be scraped).
+{:toc}
+
+
+### Core and Spark SQL
+
+ - **Major features**
+   - **Barrier Execution Mode**: [[SPARK-24374](https://issues.apache.org/jira/browse/SPARK-24374)] Support Barrier Execution Mode in the scheduler, to better integrate with deep learning frameworks.
+   - **Scala 2.12 Support**: [[SPARK-14220](https://issues.apache.org/jira/browse/SPARK-14220)] Add experimental Scala 2.12 support. Now you can build Spark with Scala 2.12 and write Spark applications in Scala 2.12.
+   - **Higher-order functions**: [[SPARK-23899](https://issues.apache.org/jira/browse/SPARK-23899)] Add a lof of new built-in functions, including high-order functions, to deal with complex data types easier.
```
kiszk: `lof` -> `lot`

ueshin: also `higher-order` instead of `high-order`?

releases/_posts/2018-11-05-spark-release-2-4-0.md
```diff
@@ -0,0 +1,188 @@
...
+
+ - **Kubernetes Scheduler Backend**
+   - [[SPARK-23984](https://issues.apache.org/jira/browse/SPARK-23984)] PySpark bindings for K8S
+   - [[SPARK-24433](https://issues.apache.org/jira/browse/SPARK-24433)] R bindings for K8S
+   - [[SPARK-23146](https://issues.apache.org/jira/browse/SPARK-23146)] Support client mode for Kubernetes cluster backend
+   - [[SPARK-23529](https://issues.apache.org/jira/browse/SPARK-23529)] Support for mounting K8S volumes
+
+ - **PySpark**
+   - [[SPARK-24215](https://issues.apache.org/jira/browse/SPARK-24215)] Implement eager evaluation for DataFrame APIs
+   - [[SPARK-22274](https://issues.apache.org/jira/browse/SPARK-22274)] User-defined aggregation functions with pandas UDF
```
kiszk: nit: `pandas` -> `Pandas`?

releases/_posts/2018-11-05-spark-release-2-4-0.md
```diff
@@ -0,0 +1,188 @@
...
+   - [[SPARK-22239](https://issues.apache.org/jira/browse/SPARK-22239)] User-defined window functions with pandas UDF
```
kiszk: ditto

releases/_posts/2018-11-05-spark-release-2-4-0.md
```diff
@@ -0,0 +1,188 @@
...
...owski, Mihaly Toth, Miklos C, Miles Yucht, Misha Dmitriev, Mukul Murthy, Mykhailo Shtelma, Neal Song, Ngone51, Nihar Sheth, Nolan Emirot, Norman Maurer, Onur Satici, Patrick McGloin, Patrick Pisciuneri, Paul Mackles, Peter Toth, Prashant Sharma, Rao Fu, Ray Burgemeestre, Rekha Joshi, Reynold Xin, Reza Safi, Ricardo Martinelli De Oliveira, Rob Vesse, Robert Kruszewski, Rong Tang, Ryan Blue, Sahil Takiar, Saisai Shao, Sandeep Singh, Sandor Murakozi, Sanket Chintapalli, Santiago Saavedra, Sean Owen, Sean Suchter, Sebastian Arzt, Shane Knapp, Shixiong Zhu, Soham Aurangabadkar, Stacy Kerkela, Stan Zhai, Stavros Kontopoulos, Steve Loughran, Sunitha Kambhampati, Takeshi Yamamuro, Takuya UESHIN, Tathagata Das, Ted Yu, Teng Peng, Thiruvasakan Paramasivan, Thomas Graves, Tom Saleeba, Vayda, Oleksandr: IT (PRG), Vinod KC, Vladimir Kuriatkov, Wang Gengliang, Weichen Xu, Wenbo Zhao, Wenchen Fan, William Sheu, XD-DENG, Xiangrui Meng, Xianjin YE, Xianyang Liu, Xiao Li, Xiaogang Tu, Xiayun Sun, Xingbo Jiang, Yacine Mazari, Yash Sharma, Ye Zhou, Yinan Li, Yogesh Garg, Yuanbo Liu, Yuanjian Li, Yuchen Huo, Yuexin Zhang, Yuming Wang, Yuval Itzchakov, Zhan Zhang, Zhang Le, Zheng RuiFeng, Zoltan C. Toth
```
gatorsmile: We need to update https://github.com/apache/spark/blob/master/dev/create-release/known_translations for the github nicknames.

kiszk:
Here is a set of output by `misspell`. I think that `Euclidian` is intentionally used while I am not familiar with ML.

```
$ spark-website > ../../bin/misspell site/docs/2.4.0/
site/docs/2.4.0/api/java/org/apache/spark/ExecutorPlugin.html:110:27: "compatability" is a misspelling of "compatibility"
site/docs/2.4.0/api/java/org/apache/spark/ml/stat/Summarizer.html:312:16: "Euclidian" is a misspelling of "Euclidean"
site/docs/2.4.0/api/java/org/apache/spark/ml/stat/Summarizer.html:345:16: "Euclidian" is a misspelling of "Euclidean"
site/docs/2.4.0/api/java/org/apache/spark/mllib/clustering/EMLDAOptimizer.html:130:34: "Teh" is a misspelling of "The"
site/docs/2.4.0/api/java/org/apache/spark/mllib/stat/MultivariateOnlineSummarizer.html:220:23: "Euclidian" is a misspelling of "Euclidean"
site/docs/2.4.0/api/java/org/apache/spark/mllib/stat/MultivariateOnlineSummarizer.html:401:23: "Euclidian" is a misspelling of "Euclidean"
site/docs/2.4.0/api/java/org/apache/spark/storage/ReadableChannelFileRegion.html:190:36: "transfered" is a misspelling of "transferred"
site/docs/2.4.0/api/python/_modules/pyspark/ml/stat.html:375:40: "Euclidian" is a misspelling of "Euclidean"
site/docs/2.4.0/api/scala/lib/scheduler.js:62:41: "existant" is a misspelling of "existent"
site/docs/2.4.0/api/scala/lib/jquery.layout.js:2312:4: "dependancy" is a misspelling of "dependency"
site/docs/2.4.0/api/scala/lib/jquery.layout.js:2629:46: "Destory" is a misspelling of "Destroy"
site/docs/2.4.0/api/scala/lib/jquery.layout.js:4767:4: "dependancies" is a misspelling of "dependencies"
site/docs/2.4.0/api/scala/lib/jquery.layout.js:4768:4: "dependancies" is a misspelling of "dependencies"
site/docs/2.4.0/api/scala/lib/jquery.layout.js:5123:4: "dependancies" is a misspelling of "dependencies"
site/docs/2.4.0/api/scala/lib/jquery.layout.js:5404:4: "dependancies" is a misspelling of "dependencies"
site/docs/2.4.0/api/scala/lib/jquery.layout.js:5434:58: "occassionally" is a misspelling of "occasionally"
site/docs/2.4.0/api/python/searchindex.js:1:138287: "composit" is a misspelling of "compost"
site/docs/2.4.0/api/python/searchindex.js:1:138596: "concret" is a misspelling of "concert"
site/docs/2.4.0/api/python/searchindex.js:1:139122: "convers" is a misspelling of "converts"
site/docs/2.4.0/api/python/searchindex.js:1:141263: "decomposit" is a misspelling of "decompose"
site/docs/2.4.0/api/python/searchindex.js:1:144690: "euclidian" is a misspelling of "euclidean"
site/docs/2.4.0/api/python/searchindex.js:1:152810: "includ" is a misspelling of "include"
site/docs/2.4.0/api/python/searchindex.js:1:163447: "occurr" is a misspelling of "occur"
site/docs/2.4.0/api/python/searchindex.js:1:166110: "portugues" is a misspelling of "portuguese"
site/docs/2.4.0/api/scala/org/apache/spark/ml/stat/Summarizer$.html:722:839: "Euclidian" is a misspelling of "Euclidean"
site/docs/2.4.0/api/scala/org/apache/spark/mllib/clustering/EMLDAOptimizer.html:55:33: "Teh" is a misspelling of "The"
site/docs/2.4.0/api/scala/org/apache/spark/mllib/clustering/package.html:173:33: "Teh" is a misspelling of "The"
site/docs/2.4.0/api/scala/org/apache/spark/mllib/stat/MultivariateOnlineSummarizer.html:533:38: "Euclidian" is a misspelling of "Euclidean"
site/docs/2.4.0/api/scala/org/apache/spark/mllib/stat/MultivariateOnlineSummarizer.html:533:133: "Euclidian" is a misspelling of "Euclidean"
site/docs/2.4.0/api/sql/css/theme.css:12:16937: "adn" is a misspelling of "and"
site/docs/2.4.0/api/sql/js/highlight.pack.js:1:11462: "lenght" is a misspelling of "length"
site/docs/2.4.0/api/sql/js/highlight.pack.js:1:92171: "seperate" is a misspelling of "separate"
site/docs/2.4.0/api/sql/js/highlight.pack.js:2:13040: "recieved" is a misspelling of "received"
site/docs/2.4.0/api/python/pyspark.ml.html:32824:16: "Euclidian" is a misspelling of "Euclidean"
site/docs/2.4.0/api/java/index-all.html:29454:23: "Euclidian" is a misspelling of "Euclidean"
$ spark-website > ../../bin/misspell release-process.md 
release-process.md:198:30: "maintance" is a misspelling of "maintenance"
```

Otherwise, we could address them if possible (not related to 2.4 release). I am neutral on them.
```
$ spark-website > ../../bin/misspell README.md _* [a-q]* releases screencasts security.md sitemap.xml sql streaming [t-z]*
news/_posts/2014-09-11-spark-1-1-0-released.md:16:451: "libary" is a misspelling of "library"
news/_posts/2016-03-09-spark-1-6-1-released.md:14:232: "signficant" is a misspelling of "significant"
releases/_posts/2014-08-05-spark-release-1-0-2.md:38:20: "consistant" is a misspelling of "consistent"
releases/_posts/2015-06-11-spark-release-1-4-0.md:28:82: "compatiblity" is a misspelling of "compatibility"
releases/_posts/2013-09-25-spark-release-0-8-0.md:16:159: "pacakge" is a misspelling of "package"
releases/_posts/2015-03-13-spark-release-1-3-0.md:19:603: "Finaly" is a misspelling of "Finally"
releases/_posts/2017-07-11-spark-release-2-2-0.md:68:13: "Availablity" is a misspelling of "Availability"
```

cloud-fan:
@kiszk thanks for checking them! Can you open a JIRA ticket to fix the typos in the doc? These typos need to be fixed in the spark repo, and they can be available in 2.4.1

For other typos that are not related to 2.4.0, feel free to open a PR to fix them. But I'm not sure if we can still update previous posts.

kiszk:
Since the above problem exists before 2.4, it is ok to fix now or to fix in another PR.

kiszk:
Got it. Since most of file are generated from Spark repo or derived from other libs, it would be good to address in other PRs. We would appreciate it if this PR could address the following long-existing problem.

```
$ spark-website > ../../bin/misspell release-process.md 
release-process.md:198:30: "maintance" is a misspelling of "maintenance"
```

cloud-fan:
@kiszk fixed at https://github.com/apache/spark-website/pull/157

srowen:
@kiszk yeah this should be a pull request against apache/spark to fix typos. That's great! I didn't know about misspell. EDIT: I just opened a PR for the rest at https://github.com/apache/spark/pull/22950

srowen (approved):

releases/_posts/2018-11-05-spark-release-2-4-0.md
cloud-fan: https://github.com/apache/spark/pull/22949

HyukjinKwon (approved):

releases/_posts/2018-11-05-spark-release-2-4-0.md
```diff
@@ -0,0 +1,188 @@
...
+ - **Spark Core**
+   - [[SPARK-25088](https://issues.apache.org/jira/browse/SPARK-25088)] Rest Server default & doc updates
+
+ - **Spark SQL**
+   - [[SPARK-23549](https://issues.apache.org/jira/browse/SPARK-23549)] Cast to timestamp when comparing timestamp with date
+   - [[SPARK-24324](https://issues.apache.org/jira/browse/SPARK-24324)] Pandas Grouped Map UDF should assign result columns by name
+   - [[SPARK-23425](https://issues.apache.org/jira/browse/SPARK-23425)] load data for hdfs file path with wildcard usage is not working properly
+   - [[SPARK-23173](https://issues.apache.org/jira/browse/SPARK-23173)] from_json can produce nulls for fields which are marked as non-nullable
+   - [[SPARK-24966](https://issues.apache.org/jira/browse/SPARK-24966)] Implement precedence rules for set operations
+   - [[SPARK-25708](https://issues.apache.org/jira/browse/SPARK-25708)] HAVING without GROUP BY should be global aggregate
```
gatorsmile: In the migration guide, we have more contents than the list of these JIRAs, right?
```
  - Since Spark 2.4, when there is a struct field in front of the IN operator before a subquery, the inner query must contain a struct field as well. In previous versions, instead, the fields of the struct were compared to the output of the inner query. Eg. if `a` is a `struct(a string, b int)`, in Spark 2.4 `a in (select (1 as a, 'a' as b) from range(1))` is a valid query, while `a in (select 1, 'a' from range(1))` is not. In previous version it was the opposite.

  - In versions 2.2.1+ and 2.3, if `spark.sql.caseSensitive` is set to true, then the `CURRENT_DATE` and `CURRENT_TIMESTAMP` functions incorrectly became case-sensitive and would resolve to columns (unless typed in lower case). In Spark 2.4 this has been fixed and the functions are no longer case-sensitive.

  - Since Spark 2.4, Spark will evaluate the set operations referenced in a query by following a precedence rule as per the SQL standard. If the order is not specified by parentheses, set operations are performed from left to right with the exception that all INTERSECT operations are performed before any UNION, EXCEPT or MINUS operations. The old behaviour of giving equal precedence to all the set operations are preserved under a newly added configuration `spark.sql.legacy.setopsPrecedence.enabled` with a default value of `false`. When this property is set to `true`, spark will evaluate the set operators from left to right as they appear in the query given no explicit ordering is enforced by usage of parenthesis.

  - Since Spark 2.4, Spark will display table description column Last Access value as UNKNOWN when the value was Jan 01 1970.

  - Since Spark 2.4, Spark maximizes the usage of a vectorized ORC reader for ORC files by default. To do that, `spark.sql.orc.impl` and `spark.sql.orc.filterPushdown` change their default values to `native` and `true` respectively.

  - In PySpark, when Arrow optimization is enabled, previously `toPandas` just failed when Arrow optimization is unable to be used whereas `createDataFrame` from Pandas DataFrame allowed the fallback to non-optimization. Now, both `toPandas` and `createDataFrame` from Pandas DataFrame allow the fallback by default, which can be switched off by `spark.sql.execution.arrow.fallback.enabled`.

  - Since Spark 2.4, writing an empty dataframe to a directory launches at least one write task, even if physically the dataframe has no partition. This introduces a small behavior change that for self-describing file formats like Parquet and Orc, Spark creates a metadata-only file in the target directory when writing a 0-partition dataframe, so that schema inference can still work if users read that directory later. The new behavior is more reasonable and more consistent regarding writing empty dataframe.

  - Since Spark 2.4, expression IDs in UDF arguments do not appear in column names. For example, a column name in Spark 2.4 is not `UDF:f(col0 AS colA#28)` but ``UDF:f(col0 AS `colA`)``.

  - Since Spark 2.4, writing a dataframe with an empty or nested empty schema using any file formats (parquet, orc, json, text, csv etc.) is not allowed. An exception is thrown when attempting to write dataframes with empty schema.

  - Since Spark 2.4, Spark compares a DATE type with a TIMESTAMP type after promotes both sides to TIMESTAMP. To set `false` to `spark.sql.legacy.compareDateTimestampInTimestamp` restores the previous behavior. This option will be removed in Spark 3.0.

  - Since Spark 2.4, creating a managed table with nonempty location is not allowed. An exception is thrown when attempting to create a managed table with nonempty location. To set `true` to `spark.sql.legacy.allowCreatingManagedTableUsingNonemptyLocation` restores the previous behavior. This option will be removed in Spark 3.0.

  - Since Spark 2.4, renaming a managed table to existing location is not allowed. An exception is thrown when attempting to rename a managed table to existing location.

  - Since Spark 2.4, the type coercion rules can automatically promote the argument types of the variadic SQL functions (e.g., IN/COALESCE) to the widest common type, no matter how the input arguments order. In prior Spark versions, the promotion could fail in some specific orders (e.g., TimestampType, IntegerType and StringType) and throw an exception.

  - Since Spark 2.4, Spark has enabled non-cascading SQL cache invalidation in addition to the traditional cache invalidation mechanism. The non-cascading cache invalidation mechanism allows users to remove a cache without impacting its dependent caches. This new cache invalidation mechanism is used in scenarios where the data of the cache to be removed is still valid, e.g., calling unpersist() on a Dataset, or dropping a temporary view. This allows users to free up memory and keep the desired caches valid at the same time.

  - In version 2.3 and earlier, Spark converts Parquet Hive tables by default but ignores table properties like `TBLPROPERTIES (parquet.compression 'NONE')`. This happens for ORC Hive table properties like `TBLPROPERTIES (orc.compress 'NONE')` in case of `spark.sql.hive.convertMetastoreOrc=true`, too. Since Spark 2.4, Spark respects Parquet/ORC specific table properties while converting Parquet/ORC Hive tables. As an example, `CREATE TABLE t(id int) STORED AS PARQUET TBLPROPERTIES (parquet.compression 'NONE')` would generate Snappy parquet files during insertion in Spark 2.3, and in Spark 2.4, the result would be uncompressed parquet files.

  - Since Spark 2.0, Spark converts Parquet Hive tables by default for better performance. Since Spark 2.4, Spark converts ORC Hive tables by default, too. It means Spark uses its own ORC support by default instead of Hive SerDe. As an example, `CREATE TABLE t(id int) STORED AS ORC` would be handled with Hive SerDe in Spark 2.3, and in Spark 2.4, it would be converted into Spark's ORC data source table and ORC vectorization would be applied. To set `false` to `spark.sql.hive.convertMetastoreOrc` restores the previous behavior.

  - In version 2.3 and earlier, CSV rows are considered as malformed if at least one column value in the row is malformed. CSV parser dropped such rows in the DROPMALFORMED mode or outputs an error in the FAILFAST mode. Since Spark 2.4, CSV row is considered as malformed only when it contains malformed column values requested from CSV datasource, other values can be ignored. As an example, CSV file contains the "id,name" header and one row "1234". In Spark 2.4, selection of the id column consists of a row with one column value 1234 but in Spark 2.3 and earlier it is empty in the DROPMALFORMED mode. To restore the previous behavior, set `spark.sql.csv.parser.columnPruning.enabled` to `false`.

  - Since Spark 2.4, File listing for compute statistics is done in parallel by default. This can be disabled by setting `spark.sql.statistics.parallelFileListingInStatsComputation.enabled` to `False`.

  - Since Spark 2.4, Metadata files (e.g. Parquet summary files) and temporary files are not counted as data files when calculating table size during Statistics computation.

  - Since Spark 2.4, empty strings are saved as quoted empty strings `""`. In version 2.3 and earlier, empty strings are equal to `null` values and do not reflect to any characters in saved CSV files. For example, the row of `"a", null, "", 1` was writted as `a,,,1`. Since Spark 2.4, the same row is saved as `a,,"",1`. To restore the previous behavior, set the CSV option `emptyValue` to empty (not quoted) string.  

  - Since Spark 2.4, The LOAD DATA command supports wildcard `?` and `*`, which match any one character, and zero or more characters, respectively. Example: `LOAD DATA INPATH '/tmp/folder*/'` or `LOAD DATA INPATH '/tmp/part-?'`. Special Characters like `space` also now work in paths. Example: `LOAD DATA INPATH '/tmp/folder name/'`.

  - In Spark version 2.3 and earlier, HAVING without GROUP BY is treated as WHERE. This means, `SELECT 1 FROM range(10) HAVING true` is executed as `SELECT 1 FROM range(10) WHERE true`  and returns 10 rows. This violates SQL standard, and has been fixed in Spark 2.4. Since Spark 2.4, HAVING without GROUP BY is treated as a global aggregate, which means `SELECT 1 FROM range(10) HAVING true` will return only one row. To restore the previous behavior, set `spark.sql.legacy.parser.havingWithoutGroupByAsWhere` to `true`.
```

cloud-fan: thanks for pointing it out! I think we don't need to mention all of them in release notes, like we don't mention all the resolved tickets. I'll pick some important ones.

js/downloads.js
```diff
@@ -21,14 +21,18 @@ var hadoop2p3 = {pretty: "Pre-built for Apache Hadoop 2.3", tag: "hadoop2.3"};
...
 // 1.4.0+
 var packagesV6 = [hadoop2p6, hadoop2p4, hadoop2p3, hadoopFree, hadoop1, cdh4, sources];
 // 2.0.0+
 var packagesV7 = [hadoop2p7, hadoop2p6, hadoop2p4, hadoop2p3, hadoopFree, sources];
 // 2.2.0+
 var packagesV8 = [hadoop2p7, hadoop2p6, hadoopFree, sources];
+// 2.4.0+
+var packagesV9 = [hadoop2p7, hadoop2p6, hadoopFree, scala2p12_hadoopFree, sources];
 
+addRelease("2.4.0", new Date("11/05/2018"), packagesV9, true, true);
```
dongjoon-hyun: Is it better to change this since the announcement is delayed? Also, file name also the old date.
```
news/_posts/2018-11-05-spark-2-4-0-released.md
```

cloud-fan: I checked 2.3.0, the date is when the vote passed. Let me update.

gatorsmile:
LGTM

Thanks! Merged to master.

PR #164, opened 2019-01-07
Suggest new Apache repo for committers

srowen:
This suggests to committers that they should use the new github remote to push to Apache.


rxin:
LGTM

PR #169, opened 2019-01-15
Add Apache Spark 2.2.3 release news and update links

dongjoon-hyun:
This PR adds or updates the followings.

**1. News**
news

**2. Release Note (No behavior change according to JIRA labels)**
release

**3. Download**
download

In addition, the release link (https://s.apache.org/spark-2.2.3) was created already for this and the mirror 2.2.2 will be removed tomorrow.


dongjoon-hyun:
Hi, @srowen .
Could you review this PR?

js/downloads.js
```diff
@@ -36,7 +36,8 @@ addRelease("2.4.0", new Date("11/02/2018"), packagesV9, true, true);
 addRelease("2.3.2", new Date("09/24/2018"), packagesV8, true, true);
 addRelease("2.3.1", new Date("06/08/2018"), packagesV8, true, true);
 addRelease("2.3.0", new Date("02/28/2018"), packagesV8, true, false);
-addRelease("2.2.2", new Date("07/02/2018"), packagesV8, true, true);
+addRelease("2.2.3", new Date("01/11/2018"), packagesV8, true, true);
```
felixcheung: 2018 -> 2019 ;)

site/js/downloads.js
```diff
@@ -36,7 +36,8 @@ addRelease("2.4.0", new Date("11/02/2018"), packagesV9, true, true);
 addRelease("2.3.2", new Date("09/24/2018"), packagesV8, true, true);
 addRelease("2.3.1", new Date("06/08/2018"), packagesV8, true, true);
 addRelease("2.3.0", new Date("02/28/2018"), packagesV8, true, false);
-addRelease("2.2.2", new Date("07/02/2018"), packagesV8, true, true);
+addRelease("2.2.3", new Date("01/11/2018"), packagesV8, true, true);
```
felixcheung: 2018 -> 2019

site/mailing-lists.html
```diff
@@ -12,7 +12,7 @@
 
   
     <meta http-equiv="refresh" content="0; url=/community.html">
-    <link rel="canonical" href="https://spark.apache.org/community.html" />
+    <link rel="canonical" href="http://localhost:4000/community.html" />
```
felixcheung: this shouldn't change?

js/downloads.js
dongjoon-hyun: oh.. Thanks.

site/mailing-lists.html
HyukjinKwon: Yea, IIRC, this happens when jeykill is ran with serve.

felixcheung (approved):

dongjoon-hyun:
Thank you, @felixcheung and @HyukjinKwon .

js/downloads.js
```diff
@@ -36,7 +36,8 @@ addRelease("2.4.0", new Date("11/02/2018"), packagesV9, true, true);
 addRelease("2.3.2", new Date("09/24/2018"), packagesV8, true, true);
 addRelease("2.3.1", new Date("06/08/2018"), packagesV8, true, true);
```
felixcheung, line 37: btw, this should be `packagesV8, true, false);` I think?

dongjoon-hyun: Yes, but `2.3.1` is still on the mirror.

felixcheung: though I tested a few mirror link for 2.3.1, seems like it is still replicated on mirrors..

felixcheung: just checked - I could remove it from dist/release later - after this is changed to false.

dongjoon-hyun: Shall I make another PR to remove this?

felixcheung: I think it's safer to put `false` on all "not-the-last" release of the minor release tree, ie. 2.3.1, 2.1.2, 2.2.2 etc. it doesn't hurt either way, for now, since 2.3.1 is still on the mirror, but it kinda should be in archive instead

srowen: That's right, and I was going to update that and remove the older releases from the mirrors just now -- I'll take care of it

PR #171, opened 2019-01-15
 Remove old releases from download page. Remove unused 'stable' flag and old info

srowen:


dongjoon-hyun:
Oh, I didn't see this PR. Could you rebase please?

srowen:
I merged this to ensure that the web page matches what is currently mirrored and in the archives. I know it's a little bit of a change to only include the latest supported releases, but think that is actually what the page should show. All remain available in the ASF archive, which is linked to from this page.

PR #187, opened 2019-04-04
Add docs for Spark 2.4.1

dbtsai:
cc @srowen , @felixcheung, @dongjoon-hyun 

I will update the rest in a followup PR. 

Thanks!


srowen (approved):
Updating the rest -- you mean the downloads page, 'latest' symlink? 
This part is pretty impossible to review but seems about right. The html files are in the right place and the file count looks right.

dbtsai:
@srowen yes, since this part is not possible to review, I decide to submit it as a separate PR. Thanks.

PR #191, opened 2019-04-04
Remove local copies of JS, CSS files we don't use directly in the site

srowen:
These aren't used anymore; copies from the internet are used or they are simply stale. The site appears fine without them too.


felixcheung (approved):
I'm all for less checked-in js

srowen:
Merged to asf-site. Yeah we shouldn't be distributing these in any way if we can avoid it, even if they aren't served, for license reasons.

PR #202, opened 2019-05-09
Spark 2.4.3: Add API doc for pyspark ML 

gatorsmile:
In the previous upgrade, the ML python APIs are not correctly generated. Now, the problem is fixed and updated.


gatorsmile:
cc @cloud-fan @srowen

cloud-fan:
thanks, merged

PR #207, opened 2019-06-18
Fix source download link

srowen:
The download javascript wasn't correctly generating the source file name. Also, minor, it wasn't correctly respecting whether it should be obtained from the mirror network or archive server.


gaborgsomogyi (approved):
LGTM.

dongjoon-hyun:
+1, late LGTM.

PR #230, opened 2019-11-07
Remove 'Stable' from download link; remove EOL 2.3.x download

srowen:
One more thing I just noticed; if we don't distinguish "Preview" releases in the download list, don't mark them all "Stable". Also, we can remove EOL 2.3.x releases from the list of current releases.

### DIFF
--- a/js/downloads.js
+++ b/js/downloads.js
@@ -18,8 +18,6 @@ var hadoop2p7 = {pretty: "Pre-built for Apache Hadoop 2.7", tag: "hadoop2.7"};
 var hadoop3p2 = {pretty: "Pre-built for Apache Hadoop 3.2 and later", tag: "hadoop3.2"};
 var scala2p12_hadoopFree = {pretty: "Pre-built with Scala 2.12 and user-provided Apache Hadoop", tag: "without-hadoop-scala-2.12"};
 
-// 2.2.0+
-var packagesV8 = [hadoop2p7, hadoop2p6, hadoopFree, sources];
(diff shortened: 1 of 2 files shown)


srowen:
@jiangxb1987

jiangxb1987 (approved):
LGTM, thanks for fixing this!

js/downloads.js
```diff
@@ -18,16 +18,13 @@ var hadoop2p7 = {pretty: "Pre-built for Apache Hadoop 2.7", tag: "hadoop2.7"};
...
-// 2.2.0+
-var packagesV8 = [hadoop2p7, hadoop2p6, hadoopFree, sources];
 // 2.4.0+
 var packagesV9 = [hadoop2p7, hadoop2p6, hadoopFree, scala2p12_hadoopFree, sources];
 // 3.0.0+
 var packagesV10 = [hadoop2p7, hadoop3p2, hadoopFree, sources];
 
 addRelease("3.0.0-preview", new Date("11/06/2019"), packagesV10, true);
 addRelease("2.4.4", new Date("08/30/2019"), packagesV9, true);
-addRelease("2.3.4", new Date("08/29/2019"), packagesV8, true);
```
HyukjinKwon, line 30: Eh, I think it's fine but I thought we'll remove this when actual 3.0.0 is released, though.

srowen: My thinking was that it's already EOL, regardless

HyukjinKwon (approved):

PR #232, opened 2019-11-22
Update the guide for adding a JIRA prefix in the test cases for other languages

HyukjinKwon:
Screen Shot 2019-11-22 at 12 31 14 PM

I was a bit hesitant about adding an actual example to the guideline as it looks a bit overwhelming  .. but I think ideally we should have such examples more in this page in the future.

Also, I was a bit wondering if this is a code style guide or pull request guide. I think it's kind of in between ..

It's discussed at http://apache-spark-developers-list.1001551.n3.nabble.com/Adding-JIRA-ID-as-the-prefix-for-the-test-case-name-td28289.html

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -305,11 +305,32 @@ Example: `Fix typos in Foo scaladoc`
 1. Consider whether documentation or tests need to be added or updated as part of the change, 
 and add them as needed.
       1. When you add tests, make sure the tests are self-descriptive.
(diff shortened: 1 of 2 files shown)


HyukjinKwon:
cc @steveloughran and @shivaram

steveloughran:
LGTM.

dongjoon-hyun (approved):
+1, LGTM. Thanks, @HyukjinKwon !

HyukjinKwon:
Merged, thanks guys.

PR #249, opened 2020-01-17
Update ml-features.html

DataWanderer:
Potential Enhancement

### DIFF
--- a/site/docs/2.4.4/ml-features.html
+++ b/site/docs/2.4.4/ml-features.html
@@ -4205,6 +4205,7 @@ for more details on the API.</p>
 
     <div class="highlight"><pre><span></span><span class="k">import</span> <span class="nn">org.apache.spark.ml.feature.ChiSqSelector</span>
 <span class="k">import</span> <span class="nn">org.apache.spark.ml.linalg.Vectors</span>
+<span class="k">import</span> <span class="nn">spark.implicits._</span>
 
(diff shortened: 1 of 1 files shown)


srowen:
You're editing the generated documentation from a previous release. We can't do that, but you can edit the docs that will generate these files in a future release. Those are in the spark repo though.
However, this doesn't explain anything about why you are changing this. Please read https://spark.apache.org/contributing.html

DataWanderer:
@srowen  The documentation is out of sync with your example code, I am just enhancing it.

srowen:
Yes, but, if so, this is not how you do it. You need to go to github.com/apache/spark/docs/...

DataWanderer:
@srowen  Why would one update the doc, when the issue in your website https://spark.apache.org/docs/latest/ml-features , also of most recent release 2.4.4

srowen:
Because the released docs are fixed, just like the bits of the release. Unless it's a big problem that makes them unusable, we fix doc problems like we fix bugs and make releases of docs and bugs.
You still haven't even described the issue, though!

DataWanderer:
These dont qualify as bugs, but wasps. Fixing wasps shouldnt need releases !

DataWanderer:
@srowen  so you are saying this is not a wasp in your repository?

srowen:
I answered you above. This is not the place to make the change.

PR #252, opened 2020-01-20
Add some info about OBJC_DISABLE_INITIALIZE_FORK_SAFETY and macOS High Sierra+ for PySpark

HyukjinKwon:
To run some PySpark features like pandas UDF, it requires to set `OBJC_DISABLE_INITIALIZE_FORK_SAFETY` on macOS High Sierra+. See https://issues.apache.org/jira/browse/SPARK-25473 and https://bugs.python.org/issue33725

I had to explain this multiple times. I wasn't particularly worried about PySpark itself because the production env is not usually Mac.

I target to describe `OBJC_DISABLE_INITIALIZE_FORK_SAFETY` in testing page as Mac is usually dev env.


dongjoon-hyun (approved):
+1, LGTM. Thank you for making this official!

HyukjinKwon:
Thanks, merged!

PR #270, opened 2020-06-18
Release Spark 3.0.0

cloud-fan:
Add the 3.0.0 [release notes](https://docs.google.com/document/d/1NrTqxf2f39AXDF8VTIch6kwD8VKPaIlLW1QvuqEcwR4/edit) and update the website.


cloud-fan:
cc @rxin @gatorsmile @dongjoon-hyun @srowen

srowen (approved):

js/downloads.js
```diff
@@ -23,7 +23,7 @@ var packagesV9 = [hadoop2p7, hadoop2p6, hadoopFree, scala2p12_hadoopFree, source
 // 3.0.0+
 var packagesV10 = [hadoop2p7, hadoop3p2, hadoopFree, sources];
 
-addRelease("3.0.0-preview2", new Date("12/23/2019"), packagesV10, true);
+addRelease("3.0.0", new Date("06/18/2020"), packagesV10, true);
```
dongjoon-hyun: The vote passed on `06/10/2020`, @cloud-fan . The date of this part is different from news release date. (cc @gatorsmile )
- https://lists.apache.org/thread.html/r1f6109df463166b826b7b2d180f74885cd95ecfaa524d722b1e665d8%40%3Cdev.spark.apache.org%3E

srowen: I don't think it should be the vote date - it wasn't available to users at that time. What you want to call the release date is a little subjective (when artifacts were uploaded? announced?) but anything like June 18 seems reasonable. It doesn't matter much.

cloud-fan: I recalled that we had a similar discussion at the time of 2.4.0 and we picked the RC vote pass date.

I've updated to use June 10.

rxin (approved):

rxin:
I find it quite strange that the news date is not today. I think we should use today’s date and then say the vote passed on June 10 and this is based on tag v3.0.0 including all commits up to June 10.

gatorsmile:
image (10)
Yes. Just checked the website, it shows a strange order.

dongjoon-hyun:
Yes. News date should be today as I mentioned here (https://github.com/apache/spark-website/pull/270#discussion_r442350319)
> The date of this part is different from news release date.

PR #272, opened 2020-06-18
Remove preview download direction.

dongjoon-hyun:
This updates the download direction because preview is removed from the dropbox.


dongjoon-hyun:
cc @cloud-fan , @gatorsmile , @srowen .

dongjoon-hyun:
We may want to delete the whole section, `Latest Preview Release`, but for the the definition and warning looks worthy to have if we have a next preview release someday. So, I only remove the invalid part.

srowen (approved):

dongjoon-hyun:
Thank you, @srowen .

PR #281, opened 2020-08-03
Update committer page

mridulm:
I joined LinkedIn

### DIFF
--- a/committers.md
+++ b/committers.md
@@ -55,7 +55,7 @@ navigation:
 |Jungtaek Lim|Cloudera|
 |Sean McNamara|Oracle|
 |Xiangrui Meng|Databricks|
-|Mridul Muralidharam|Google|
+|Mridul Muralidharam|LinkedIn|
 |Andrew Or|Princeton University|
 |Kay Ousterhout|LightStep|
 |Sean Owen|Databricks|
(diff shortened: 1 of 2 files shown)


srowen (approved):

mridulm:
@srowen I had removed local changes to site/developer-tools.html, site/examples.html, site/security.html, site/sitemap.xml made during to jekyl build before submitting pr - but looks like a merge will regenerate them ...
Is there a way to suppress this ? 

Looks like version changes are triggering bunch of whitespace related mods - which will keep toggling based on which version everyone is on.

I am on 3.8.6 of jekyll (new box, so most deps are new from apt)

HyukjinKwon (approved):
:)

HyukjinKwon:
@mridulm, yeah they're some differences per the Jekyll version. I tried to make some fixes to minimise the diff before but not sure if we can have a great way to avoid them. I usually do the same thing as you did.

gengliangwang (approved):

srowen:
I'm on jekyll 4.1.0, and ideally, we'd use the latest one consistently. But yeah I usually revert any spurious changes and don't deal with it

mridulm:
Sounds good, thanks @HyukjinKwon, @srowen ! Merged

PR #283, opened 2020-08-10
Fix Python 2 examples of Pi Estimation

KoyamaSohei:
I tried to run [Pi Estimation example](https://spark.apache.org/examples.html)  with Python 3, but it is not working.

According to [latest docs](https://spark.apache.org/docs/latest/) ,
> Python 2 and Python 3 prior to version 3.6 support is deprecated as of Spark 3.0.0.

[Basic example](https://github.com/apache/spark/tree/master/examples/src/main/python) seemed to be fixed already.

### DIFF
--- a/examples.md
+++ b/examples.md
@@ -88,9 +88,9 @@ def inside(p):
     x, y = random.random(), random.random()
     return x*x + y*y < 1
 
-count = sc.parallelize(xrange(0, NUM_SAMPLES)) \
+count = sc.parallelize(range(0, NUM_SAMPLES)) \
              .filter(inside).count()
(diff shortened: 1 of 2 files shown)


HyukjinKwon:
@KoyamaSohei, the change looks good. Can you double check if there are any other examples to fix? It should be good to fix them in batch while we're here.

KoyamaSohei:
I'll try to check if all the samples in https://spark.apache.org/examples.html to work with Python 3, thanks

HyukjinKwon (approved):

KoyamaSohei:
As far as I can tell, there doesn't seem to be any fixes related to Python 2.

srowen:
Merged to asf-site, thank you.

PR #312, opened 2021-02-26
[SPARK-34539][DOC] Cleanup Zinc description in developer-tools

Yikun:
Since Spark 3.0.0 with scala-maven-plugin v4.x, if using `build/mvn`, by default, the Zinc incremental compiler is enabled by maven plugin `scala-maven-plugin`, it will call directly with embedded Zinc APIs to speed up your compilation.

There is no longer something developers need to care about, so this patch we cleanup all Zinc related description.

Releted: https://github.com/apache/spark/pull/31647

### DIFF
--- a/developer-tools.md
+++ b/developer-tools.md
@@ -34,20 +34,6 @@ $ ./bin/spark-shell
 $ build/sbt ~compile
 ```
 
-<h4>Maven: Speeding up Compilation with Zinc</h4>
-
-[Zinc](https://github.com/typesafehub/zinc) is a long-running server version of SBT's incremental
-compiler. When run locally as a background process, it speeds up builds of Scala-based projects
-like Spark. Developers who regularly recompile Spark with Maven will be the most interested in
-Zinc. The project site gives instructions for building and running `zinc`; OS X users can
-install it using `brew install zinc`.
-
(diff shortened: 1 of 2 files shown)


HyukjinKwon:
@Yikun the PR should also contain generated HTMLs via running `bundle exec jekyll build`.

Yikun:
@HyukjinKwon Yes, thanks for your reminder, and would you mind give some suggestion on the doc? I will adress all problems in next PS.

srowen (commented):
This is OK. But we can also just delete all of this, as it's no longer something developers need to care about as it just happens.

Yikun:
@srowen yep, sounds reasonable.

And there are only one reason to keep it left, from this doc, the developer can undertand the mechanism of scala-maven-plugin with Zinc and tell user can use `recompileMode` configuration to enable the incremental comiler or not.

But it's also ok for me to cleanup these doc. Do we need more inputs from others? : )

srowen:
I would say just delete it, unless anyone else chimes in now with an objection.

Yikun:
@srowen OK, thanks for your suggestion, done. and I also delete the description in Spark doc.

[1] https://github.com/apache/spark/pull/31647/commits/27b94a3b2da7913558a37e9678b48cbdac306717

PR #313, opened 2021-02-26
[SPARK-34539][DOC] Cleanup Zinc security issue

Yikun:
Zinc Server realted code has been cleanup [1], this patch
removes the Zinc Server related security issue.

[1] https://github.com/apache/spark/pull/31647

### DIFF
--- a/security.md
+++ b/security.md
@@ -135,38 +135,6 @@ Credit:
 - Andre Protas, Apple Information Security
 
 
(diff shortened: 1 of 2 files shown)


srowen:
No, don't delete this. It's still important historical info even though it was indeed already mitigated, and not an issue in active branches.

Yikun:
@srowen OK, thanks for explanation.

PR #317, opened 2021-02-27
Add a SQL Metrics section to the sql documentation.

sririshindra:
This [PR](https://github.com/apache/spark/pull/31477) adds a new metric to the SQL UI named "number of matched rows". During the review process @maropu pointed out that there is no documentation explaining the metrics available in the spark UI and we should probably add that to the spark website so that user can understand what exactly does the metrics mean. This PR is an attempt to add the documentation for the some of the metrics available in the SQL Tab of the Spark UI. There might be more metrics that can be further added to this new section that I added.

image


maropu:
Could you make a PR for the spark codebase `https://github.com/apache/spark/tree/master/docs`? The documents are automatically generated from there.

sririshindra:
> Could you make a PR for the spark codebase `https://github.com/apache/spark/tree/master/docs`? The documents are automatically generated from there.

Sorry, I didn't realize that earlier. I will make a PR for the spark code base as you suggested. Thanks.

PR #318, opened 2021-02-27
Fix a sbt example for generating dependency graphs

maropu:
This PR intends to fix an error below;
```
$./build/sbt dependency-tree
[error] Not a valid command: dependency-tree
[error] Not a valid project ID: dependency-tree
[error] Expected ':'
[error] Not a valid key: dependency-tree (similar: dependencyTree, dependencyOverrides, sbtDependency)
[error] dependency-tree
[error]                ^
```

### DIFF
--- a/developer-tools.md
+++ b/developer-tools.md
@@ -374,7 +374,7 @@ $ git checkout origin/pr/112 -b new-branch
 
 ```
 $ # sbt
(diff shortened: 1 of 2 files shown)


maropu:
cc: @srowen

srowen (approved):

PR #332, opened 2021-04-16
Add error message guidelines

karenfeng:
Adds error message guidelines to the contributing guidelines, as [discussed in the developers list](http://apache-spark-developers-list.1001551.n3.nabble.com/DISCUSS-Build-error-message-guideline-td31076.html).

For brevity, I included a condensed version of the error message guideline on the main contributing page, and linked to the full guideline for more detail.

Main contributing page:

contributing

Full guideline:

error_message_guidelines

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -202,10 +202,10 @@ When writing error messages, you should:
 - Use active voice
 - Avoid time-based statements, such as promises of future support
 - Use the present tense to describe the error and provide suggestions
-- Provide concrete examples
+- Provide concrete examples if the resolution is unclear
 - Avoid sounding accusatory, judgmental, or insulting
 - Be direct
-- Use human-readable text for user-facing errors
+- Do not use programming jargon in user-facing errors
 
 See the <a href="{{site.baseurl}}/error-message-guidelines.html">error message guidelines</a> for more details.
 
--- a/contributing.md
+++ b/contributing.md
@@ -186,6 +186,29 @@ that maintainability, consistency, and code quality come first. New algorithms s
 - Come with a reasonable expectation of developer support.
 - Have `@Since` annotation on public classes, methods, and variables.
 
+<h3>Error Message Guidelines</h3>
+
+Exceptions thrown in Spark should be associated with standardized and actionable
+error messages.
+
+Error messages should answer the following questions:
+
+- **What** was the problem?
+- **Why** did the problem happen?
+- **How** can the problem be solved?
+
+When writing error messages, you should:
+
+- Use active voice
+- Avoid time-based statements, such as promises of future support
+- Use the present tense to describe the error and provide suggestions
+- Provide concrete examples
+- Avoid sounding accusatory, judgmental, or insulting
+- Be direct
+- Use human-readable text for user-facing errors
+
+See the <a href="{{site.baseurl}}/error-message-guidelines.html">error message guidelines</a> for more details.
+
 <h3>Code Review Criteria</h3>
 
 Before considering how to contribute code, it's useful to understand how code is reviewed, 
--- a/error-message-guidelines.md
+++ b/error-message-guidelines.md
@@ -19,6 +19,7 @@ Exceptions thrown from Spark should answer the Five W's and How:
 -   **Who** encountered the problem?
 -   **What** was the problem?
 -   **When** did the problem happen?
+-   **Where** did the problem happen?
 -   **Why** did the problem happen?
 -   **How** can the problem be solved?
 
@@ -45,14 +46,15 @@ In many cases, the error message should explicitly answer **what**,
 </code>
 
 -   **What:** Unable to generate encoder inner class.
--   **Why:** Did not have access to the scope that the class was defined
-    in.
+-   **Why:** Did not have access to the scope that the class was defined in.
 -   **How:** Try moving this class out of its parent class.
 
 ##### Example 2
 
 If the proposed fix (**how**) feels arbitrary, providing an explanation
 for **why** the error occurred can reduce user frustration.
+
+**Before**
 
 <code style="white-space:normal">
   <a href="https://github.com/apache/spark/blob/03dd33cc982ebb3de4354274ac49da31521b8195/sql/catalyst/src/main/scala/org/apache/spark/sql/errors/QueryCompilationErrors.scala#L498">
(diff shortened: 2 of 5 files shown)


HyukjinKwon (approved):

karenfeng:
Right now, the error message guidelines page is not accessible via the main navigation bar, and is only accessible via a link from the [Contributing to Spark](https://spark.apache.org/contributing.html) page. Is there a precedent for semi-orphan pages?

HyukjinKwon:
I think it's fine.

wangyum (approved):

wangyum:
cc @cloud-fan @gatorsmile @dongjoon-hyun

gatorsmile:
@karenfeng Could you post the screen shot in the PR description?

karenfeng:
Good idea, @gatorsmile! Posted the screenshots for both the contributing page (the condensed error message guidelines are in the middle) and the full error message guidelines page.

error-message-guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+  <a href="https://github.com/apache/spark/blob/03dd33cc982ebb3de4354274ac49da31521b8195/sql/catalyst/src/main/scala/org/apache/spark/sql/errors/QueryCompilationErrors.scala#L498">
+    <span style="background-color: #ffcccc">Unsupported</span> function name {}.
+  </a>
+</code>
+
+-   **What:** Unsupported function name.
+-   **Why:** Unclear.
+-   **How:** Unclear.
+
+**After**
```
dongjoon-hyun, line 69: It seems that we need to add a matching `**Before**`.

error-message-guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+catalog. Specify a function name with one or two parts.</span>*
+
+-   **What:** Invalid function name.
+-   **Why:** Temporary functions cannot belong to a catalog.
+-   **How:** Specify a function name with one or two parts.
+
+#### Implicitly answer How
+
+Not all error messages should be this verbose. Sometimes, explicitly
+explaining **how** to resolve the problem would be redundant.
```
dongjoon-hyun, line 79: In this context, `redundant` -> `sufficient`?

error-message-guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+
+**After**
+
+<code style="white-space:normal">
+  Invalid decimal {}; <span style="background-color: #ccffcc">encountered error while parsing at position {}.</span>
+</code>
+
+-   **What:** Invalid decimal.
+-   **Why**: The decimal parser encountered an error at the specified
+    position.
```
dongjoon-hyun: Like line 143, shall we move this `position` to line 141?

dongjoon-hyun (commented):
Thank you, @karenfeng . This is a nice document.

However, as a guideline document, several corner case descriptions seem to increase the ambiguity. For example, what is recommend by the following sentence and the word `redundant`?
> Sometimes, explicitly explaining **how** to resolve the problem would be redundant.

I'm wondering if you want to say that `you may skip HOW in this case` or `you should not add redundant `HOW` in the error message. Since this is a guideline, could you be more specific about our position?

contributing.md
```diff
@@ -186,6 +186,29 @@ that maintainability, consistency, and code quality come first. New algorithms s
...
+- **Why** did the problem happen?
+- **How** can the problem be solved?
+
+When writing error messages, you should:
+
+- Use active voice
+- Avoid time-based statements, such as promises of future support
+- Use the present tense to describe the error and provide suggestions
+- Provide concrete examples
+- Avoid sounding accusatory, judgmental, or insulting
```
dongjoon-hyun, line 206: Just a question. Did we have the error message in this category before? I'm wondering why `You must` is categorized into this category.
> You must specify an amount for {}

contributing.md
```diff
@@ -186,6 +186,29 @@ that maintainability, consistency, and code quality come first. New algorithms s
...
+- **What** was the problem?
+- **Why** did the problem happen?
+- **How** can the problem be solved?
+
+When writing error messages, you should:
+
+- Use active voice
+- Avoid time-based statements, such as promises of future support
+- Use the present tense to describe the error and provide suggestions
+- Provide concrete examples
```
dongjoon-hyun: Oh, are we enforcing to have `For example, ...` clause always?

contributing.md
```diff
@@ -186,6 +186,29 @@ that maintainability, consistency, and code quality come first. New algorithms s
...
+- Avoid sounding accusatory, judgmental, or insulting
+- Be direct
+- Use human-readable text for user-facing errors
```
dongjoon-hyun: Although I understand the meaning, I'd like to recommend to revise this guideline.

Given the example, this guideline aims to discourage `!=` and to recommend a natural sentence. Technically, both are human-readable text, aren't they?
```
- RENAME TABLE source and destination databases do not match: '{}' != '{}'
- RENAME TABLE source and destination databases do not match. The source database is {}, but the destination database is {}.
```

error-message-guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+---
+
+## Error Message Guidelines
+
+This guide is a reference for composing standardized and actionable
+error messages in Apache Spark.
+
+### Include What, Why, and How
+
+Exceptions thrown from Spark should answer the Five W's and How:
```
dongjoon-hyun, line 17: Although we know `where` means the stack trace, please add `**Where**` in the following list. Currently, we say `the Five W's` in this sentence and show only four Ws in the following list. So, this mismatch unnecessarily confuse readers.

error-message-guidelines.md
karenfeng: Good catch, thanks!

contributing.md
karenfeng: To avoid making error messages unnecessarily verbose, we should likely only encourage developers to add examples in the case that the fix is unclear otherwise. I can clarify the wording here.

error-message-guidelines.md
karenfeng: Whoops, good catch. Thanks!

contributing.md
karenfeng: "Must" sounds more like a command, so it may put users off - we don't want them to feel like they did something wrong. Luckily, we don't have many good examples of more accusatory/judgmental/insulting error messages.

contributing.md
karenfeng: You're right, human-readable is likely the wrong phrase in this case. A less expansive suggestion could be: "Do not use programming jargon in user-facing errors"

karenfeng:
Thanks for the review @dongjoon-hyun! Let me know if I missed any comments, or if you have any further improvements to suggest.

dongjoon-hyun (approved):
+1, LGTM. Thank you so much, @karenfeng .

PR #335, opened 2021-05-04
Update contributing to include code of conduct section

dennyglee:
Updates to contributing.md and contributing.html to include call out to ASF code of conduct.


HeartSaVioR:
I think the rationalization is great, but why not going through dev@ mailing list? Many contributors are subscribing dev@ mailing list as well and it would be also a good time to remind the CoC from your idea/discussion thread.

I assume getting consensus to add here is just a matter of time (as CoC is already something ASF project requires to us, and we just make it explicit here), but might be ideal to reach more audiences.

srowen:
I sent to dev@ too for visibility. This seems fine.

PR #338, opened 2021-05-11
Update the messages for the new repository service

xuanyuanking:
In SparkSubmit, we use the same environment variable `DEFAULT_ARTIFACT_REPOSITORY` for both central and spark-packages. So it will cause other problems if setting It to `https://repos.spark-packages.org/`.

In this PR we change the workaround to manually download the artifacts.


cloud-fan (approved):

HyukjinKwon (approved):

HyukjinKwon:
Merged

PR #355, opened 2021-08-18
Use ASF mail archives not defunct nabble links

srowen:
Nabble archive links appear to not work anymore. Use ASF pony mail links instead for archives.


zero323:
I was looking into it and [the following](https://support.nabble.com/Downsizing-Nabble-td7609715.html):

> Forum owners who want their forum preserved can post to this support forum to let us know to move their forum to that one server.  Note that we no longer do mailing list archiving, so if you own an old mailing list archive, there is no point to preserving it. 

So it is seems like it definitely not going back in our case.

zero323 (approved):

HyukjinKwon:
lgtm2

PR #357, opened 2021-10-08
Remove duplicated command instruction from developer-tools

yutoacts:
I'm not so sure if it is intended but `build/sbt compile` is duplicated in developer-tools.


srowen:
I think it could be on purpose, to illustrate 'iterating', but doesn't matter either way, can be removed

PR #359, opened 2021-10-11
New home page and layout for Spark website

gengliangwang:
This PR introduces a new UI for the Spark Website, including:
* A new home page
* A new global layout template
* Update title case capitalization of each page

To preview the new website, check http://ec2-54-184-63-212.us-west-2.compute.amazonaws.com


gengliangwang:
cc @mateiz @rxin @gatorsmile @srowen @cloud-fan @HyukjinKwon @dongjoon-hyun @holdenk

HyukjinKwon (approved):
I reviewed offline before this PR. LGTM from my side but would better to have a second look

zero323:
Looks nice, but I am bit concerned about the contrast ratio. If we had

```css
body {
    color: #555555;
    ...
}
```

it should still look good, IMHO, but exceed WCAG AAA on normal text (going from 5.74:1 to 7.45:1).

gengliangwang:
@zero323 Thanks for the suggestion. Even though it fails WCAG AAA, but it still passes WCAG AA. I prefer the current color (#666666) since it looks more comfortable on my first impression.

I made two screenshots. If there are more suggestions to change the color, I will update it.
`#666666`
image

`#555555`
image

dongjoon-hyun (approved):

dongjoon-hyun:
New look is very impressive, @gengliangwang . :)

holdenk (approved):
I like the new design, thanks for working on this :)

I think there are maybe some things we could follow up on (e.g. only Python has install instructions on the main page), but we can leave the fine-tuning until post-merge this is definitely an improvement over what we have now

cloud-fan (approved):

gatorsmile (approved):
LGTM. Thanks!

mateiz:
This looks great to me, very excited about it. We should definitely keep improving it over time, for instance with better code examples and setup instructions in all the languages if applicable. My only comment on that is to make sure to keep the first example people see small. The other tabs can contain more stuff.

yaooqinn (approved):

yaooqinn:
LGTM

gengliangwang:
Merged. Thanks all for the review!

HeartSaVioR:
Late LGTM. Looks nice!

PR #363, opened 2021-10-19
Fix Spark 3.2.0 download URL

gengliangwang:
I made a mistake on the download URL of Spark 3.2.0. This PR is to fix it.

### DIFF
--- a/js/downloads.js
+++ b/js/downloads.js
@@ -15,7 +15,7 @@ var sources = {pretty: "Source Code", tag: "sources"};
 var hadoopFree = {pretty: "Pre-built with user-provided Apache Hadoop", tag: "without-hadoop"};
 var hadoop2p7 = {pretty: "Pre-built for Apache Hadoop 2.7", tag: "hadoop2.7"};
 var hadoop3p2 = {pretty: "Pre-built for Apache Hadoop 3.2 and later", tag: "hadoop3.2"};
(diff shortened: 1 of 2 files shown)


cloud-fan (approved):

cloud-fan:
We should really rename the maven profile name...

yaooqinn:
> We should really rename the maven profile name...

😂

gengliangwang:
There are some discussions about this on dev list: https://www.mail-archive.com/dev@spark.apache.org/msg28025.html
The conclusion was to make it at Spark 3.3.0.  
We could try renaming the tarball on finalizing the release, but I was using release script...
Should we keep it as it is? Or figure out a way to fix it? cc @dongjoon-hyun @sunchao

srowen:
Change at Spark 3.3 IMHO. It's just a naming thing. hadoop-3.2 really means "3.2 or later" so it's not crazy

sunchao:
I'll pick up https://github.com/apache/spark/pull/30891 again soon and make sure we'll get this done before Spark 3.3.0

> Should we keep it as it is? Or figure out a way to fix it? 

I think we can fix this in the same PR above, perhaps by just changing `make-distribution.sh`.

srowen:
Let's not make the release differ from the profiles, I think. We can rename the profile and leave hadoop-3.2 as a no-op profile. By 3.3.0, who knows, maybe we're on 3.4 or something

dongjoon-hyun:
Thank you for the fix, @gengliangwang .

PR #377, opened 2022-01-28
fix display problem

huaxingao:
I think this `|` inside ` Support (IGNORE | RESPECT) ...` caused display problem. The dot looks strange and the `|` in between of `IGNORE` and `RESPECT` is gone. 
```
- [[SPARK-30789]](https://issues.apache.org/jira/browse/SPARK-30789): Support (IGNORE | RESPECT) NULLS for LEAD/LAG/NTH_VALUE/FIRST_VALUE/LAST_VALUE
```
Screen Shot 2022-01-28 at 2 39 37 PM

I will also fix the spaces in this file. All the other release md files have two spaces before each of the line for the jira list, but my intelliJ setting is not right, all the spaces are gone so I will manually add these spaces back.

### DIFF
--- a/releases/_posts/2022-01-26-spark-release-3-2-1.md
+++ b/releases/_posts/2022-01-26-spark-release-3-2-1.md
@@ -15,30 +15,30 @@ Spark 3.2.1 is a maintenance release containing stability fixes. This release is
 
 ### Notable changes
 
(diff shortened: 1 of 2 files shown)


huaxingao:
cc @srowen @dongjoon-hyun

dongjoon-hyun (approved):

srowen (approved):

huaxingao:
Merged. Thanks @srowen @dongjoon-hyun

PR #387, opened 2022-05-10
Add SQL API homepage into sitemap.xml

gengliangwang:
Currently, Google search shows outdated result for Spark SQL functions. For example, searching `date_trunc spark` will show the result https://spark.apache.org/docs/2.3.0/api/sql/index.html. Note that the example of `date_trunc` is wrong. The latest example is correct but it is not shown in google search.

This doesn't happen to other APIs. For example, search `spark rdd scala api` will show the latest API page.

I think the problem is on the sitemap.xml. We should include SQL API homepage.

### DIFF
--- a/site/sitemap.xml
+++ b/site/sitemap.xml
@@ -137,6 +137,11 @@
   <changefreq>weekly</changefreq>
   <priority>1.0</priority>
 </url>
+<url>
+  <loc>https://spark.apache.org/docs/latest/api/sql/index.html</loc>
+  <changefreq>weekly</changefreq>
+  <priority>1.0</priority>
+</url>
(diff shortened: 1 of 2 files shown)


yaooqinn (approved):

viirya (approved):

cloud-fan (approved):

HyukjinKwon (approved):

gengliangwang:
@yaooqinn @viirya @cloud-fan @HyukjinKwon Thanks for the review.  I will merge this one and see if the search result is updated within one week.

dongjoon-hyun (commented):
+1, LGTM

PR #410, opened 2022-08-04
Change some <h4> tags in last commit to <h3>.

MacrothT:
Change level of HTML tags to match original Markdown headings that were ###.


srowen:
Oh, BTW, have you fixed the source markdown in the Spark distro? I forgot to ask. Normally we don't edit old releases' generated docs unless it's just not working

MacrothT:
Thanks for noticing, Sean. After failing in syncing with the repo, I was thinking of updating html first.
Still no luck today. Will see what I can do with that later.

PR #412, opened 2022-09-08
correct typo in rdd-programming-guide

abubakrsiddq:
Fraction was repeated twice in guide corrected it

Also corrected in  spark docs in this PR https://github.com/apache/spark/pull/37839

### DIFF
--- a/site/docs/2.2.0/rdd-programming-guide.html
+++ b/site/docs/2.2.0/rdd-programming-guide.html
@@ -1088,7 +1088,7 @@ for details.</p>
 </tr>
 <tr>
   <td> <b>sample</b>(<i>withReplacement</i>, <i>fraction</i>, <i>seed</i>) </td>
-  <td> Sample a fraction <i>fraction</i> of the data, with or without replacement, using a given random number generator seed. </td>
+  <td> Sample a <i>fraction</i> of the data, with or without replacement, using a given random number generator seed. </td>
 </tr>
 <tr>
   <td> <b>union</b>(<i>otherDataset</i>) </td>


site/docs/2.2.0/rdd-programming-guide.html
```diff
@@ -1088,7 +1088,7 @@ <h3 id="transformations">Transformations</h3>
 </tr>
 <tr>
   <td> <b>sample</b>(<i>withReplacement</i>, <i>fraction</i>, <i>seed</i>) </td>
-  <td> Sample a fraction <i>fraction</i> of the data, with or without replacement, using a given random number generator seed. </td>
+  <td> Sample a <i>fraction</i> of the data, with or without replacement, using a given random number generator seed. </td>
```
srowen, line 1091: This isn't a typo. It's saying that the fraction is given by the argument _fraction_. 
Also, this is a PR against a copy of the docs generated for Spark 2.2.0. You would fix docs in the apache/spark source tree for the next release.

srowen:
You're right, but, that's a copy of the documentation from a very old version.

site/docs/2.2.0/rdd-programming-guide.html
abubakrsiddq: args are specified in the left field. looks to be a typo 
image
 Also do we just correct this file then https://github.com/apache/spark/blob/master/docs/rdd-programming-guide.md even for the site?

abubakrsiddq: i was reading contribution gyuideline and it said to updatate here as well

srowen: Yes you would update that .md file, but I do not think this is a typo.

abubakrsiddq: I felt that the italics was enough for giving away its an arg, repetition would not be right.

PR #414, opened 2022-10-03
CVE version update

srowen:
See mailing list discussion. The idea is to give a 'resolved by' version for older CVEs that are advice or affected only the build.


dongjoon-hyun (approved):
+1, LGTM. Thank you!

viirya (approved):

PR #423, opened 2022-10-26
Auto update the copyright year 

wangyum:


srowen (commented):
Let's just take out the year? I don't think it has any value or significance

wangyum:
OK. close it.

srowen:
I was thinking to remove the year rather than update but either way, yes it doesn't matter to copyright

wangyum:
Removed the copyright year.

srowen:
I think it's fine. I think it doesn't matter either way

wangyum:
I'm prefer keep the year:
image

PR #425, opened 2022-11-01
Remove Spark 3.1.x from downloads as it is EOL

srowen:


dongjoon-hyun (commented):
+1, late LGTM. Thank you so much.

PR #461, opened 2023-05-10
Update Apache Spark 3.5 Release Window

xinrong-meng:
Update Apache Spark 3.5 Release Window, with proposed dates:

```
| July 16th 2023 | Code freeze. Release branch cut.|
| Late July 2023 | QA period. Focus on bug fixes, tests, stability and docs. Generally, no new features merged.|
| August 2023    | Release candidates (RC), voting, etc. until final release passes|
```

### DIFF
--- a/site/versioning-policy.html
+++ b/site/versioning-policy.html
@@ -250,7 +250,7 @@ available APIs.</p>
 Hence, Spark 2.3.0 would generally be released about 6 months after 2.2.0. Maintenance releases happen as needed
 in between feature releases. Major releases do not happen according to a fixed schedule.</p>
 
-<h3>Spark 3.4 release window</h3>
+<h3>Spark 3.5 release window</h3>
 
 <table>
   <thead>
@@ -261,15 +261,15 @@ in between feature releases. Major releases do not happen according to a fixed s
   </thead>
   <tbody>
     <tr>
-      <td>January 16th 2023</td>
+      <td>July 16th 2023</td>
       <td>Code freeze. Release branch cut.</td>
     </tr>
     <tr>
-      <td>Late January 2023</td>
+      <td>Late July 2023</td>
       <td>QA period. Focus on bug fixes, tests, stability and docs. Generally, no new features merged.</td>
     </tr>
     <tr>
-      <td>February 2023</td>
+      <td>August 2023</td>
       <td>Release candidates (RC), voting, etc. until final release passes</td>
     </tr>
   </tbody>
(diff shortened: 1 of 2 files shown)


site/versioning-policy.html
```diff
@@ -261,15 +261,15 @@ <h3>Spark 3.4 release window</h3>
   </thead>
   <tbody>
     <tr>
-      <td>January 16th 2023</td>
+      <td>July 16th 2023</td>
```
srowen, line 264: This feels too soon? 3.4 was released a month ago. This schedule should be at least planned for +2 months later than this says, I think

HyukjinKwon: Actually I think this was discussed at https://lists.apache.org/thread/l6tprb580lrq1118c83kt9fst30jp8n3 . 3.4 was a bit delayed .. but might be good to stick to the plan ..(?)

srowen: OK, not against it. The real cadence seems like it's 8 months, though, we could still start the process 'on schedule' and just plan for it taking longer. We'll have a release here with 4 months or so of changes. If the next one slips, we might see releases with just 3 or 2 months. 6 months might be just ambitious

dongjoon-hyun: I'm good with this PR.

BTW, it seems that we are going to start 3.5.0 and 3.4.1 release at the same time.

xinrong-meng: By convention, do we avoid releasing two versions simultaneously?

dongjoon-hyun: No~ @xinrong-meng  It's not an issue at all.

xinrong-meng: Thanks @dongjoon-hyun !

Given that we proposed to establish a predictable release cadence early this year, it may be beneficial to stick to that plan for the next release. However, if 3.5.0 experiences further delays, it could be a strong indication that we should reconsider the feasibility of the current release cadence and propose a less ambitious schedule going forward. WDYT @srowen

srowen: Yeah OK. Another argument is that with only a few months of changes that the release process should hit fewer surprises

xinrong-meng: Totally agree! :)

dongjoon-hyun: Ya, that's really true.

HyukjinKwon (approved):

dongjoon-hyun (approved):

yaooqinn (approved):

yaooqinn:
Shall we mail dev like 3.4/3.3/3.2/3.1...?

https://lists.apache.org/list?dev@spark.apache.org:lte=3y:release%20window

xinrong-meng:
@yaooqinn Certainly, I'll send an email now!

yaooqinn:
thanks @xinrong-meng

PR #527, opened 2024-07-01
Add xDiscover to third-party tools list

azadmajd:
Change the builder version and ruby version. the pipeline was broken last time according to mismatched versioning of these 2


srowen:
Please revert the unrelated Gemfile changes, otherwise this is OK

PR #532, opened 2024-07-15
Update developer-tools.md

computerscienceiscool:
Fixed a couple of spelling errors


srowen:
Looks good, you just need to regenerate the corresponding .html file with jekyll (notes on how to do it are in the project README). Or, for a simple change you could carefully also edit the .html file

PR #578, opened 2025-02-06
Self-host JavaScript and CSS in Spark 3.5.4 website

gengliangwang:
### What changes were proposed in this pull request?

Currently, the [Spark website](https://spark.apache.org/docs/latest/) fails to load bootstrap.bundle.min.js/jquery.js/etc due to the following errors

```

Refused to load the stylesheet '<URL>' because it violates the following Content Security Policy directive: "style-src 'self' 'unsafe-inline' data:". Note that 'style-src-elem' was not explicitly set, so 'style-src' is used as a fallback.
Understand this errorAI
latest/:1 Refused to load the script 'https://cdn.jsdelivr.net/npm/bootstrap@5.0.2/dist/js/bootstrap.bundle.min.js' because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-inline' 'unsafe-eval' https://analytics.apache.org/". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.
Understand this errorAI
latest/:1 Refused to load the script 'https://code.jquery.com/jquery.js' because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-inline' 'unsafe-eval' https://analytics.apache.org/". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.
...
```

As a result, the website looks like:
image

To fix the issue, I suggest using self-host javascript and CSS in Spark website

### Why are the changes needed?

Fix the broken styling on the Spark website.

### Does this PR introduce _any_ user-facing change?

No

### How was this patch tested?

Test locally

### Was this patch authored or co-authored using generative AI tooling?

No


ueshin (approved):

gengliangwang:
I will merge this one and see if it works.

PR #582, opened 2025-02-10
Add an example for get_json_object when the JSON object is of JSON array type

fusheng9399:
Most users are unaware of how to retrieve the corresponding JSON object using get_json_object when the JSON object is of a JSON array type. 
Before this PR,  the example:
```java
SELECT get_json_object('{"a":"b"}', '$.a');
+-------------------------------+
|get_json_object({"a":"b"}, $.a)|
+-------------------------------+
|                              b|
+-------------------------------+
``` 

After this PR, the example of get_json_object has been changed to look like this:
```java
SELECT get_json_object('{"a":"b"}', '$.a');
+-------------------------------+
|get_json_object({"a":"b"}, $.a)|
+-------------------------------+
|                              b|
+-------------------------------+

SELECT get_json_object('[{"a":"b"},{"a":"c"}]', '$[*].a');
+----------------------------------------------+
|get_json_object([{"a":"b"},{"a":"c"}], $[*].a)|
+----------------------------------------------+
|                                     ["b","c"]|
+----------------------------------------------+
```

### DIFF
--- a/site/docs/4.0.0-preview2/generated-json-funcs-examples.html
+++ b/site/docs/4.0.0-preview2/generated-json-funcs-examples.html
@@ -28,6 +28,13 @@
 <span class="o">|</span><span class="w">                              </span><span class="n">b</span><span class="o">|</span>
 <span class="o">+</span><span class="c1">-------------------------------+</span>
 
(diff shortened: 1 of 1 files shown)


fusheng9399:
cc @dongjoon-hyun @zhengruifeng @LuciferYang @HyukjinKwon

LuciferYang (changes requested):
No, we can't directly modify the HTML files here. We should add examples in the Spark codebase, and they will be updated after new version release.

PR #587, opened 2025-02-16
The ASF storefront no longer exists. Updated the Shop section to reflect it and also removed an unneeded image as a result.

james-malone:


srowen (changes requested):
Good catch. I think I'd just remove this section entirely, if there's not a shop to link to now, and can add a new one later.

Also can you regenerate the HTML? it has to be changed too to update the site. jekyll does it, but you could also easiliy just edit the HTML by hand to remove the section

srowen:
Closed in favor of https://github.com/apache/spark-website/pull/603

PR #590, opened 2025-02-21
Security

raboof:
The spark security page already links to the Apache-wide security information. This change removes the link from the menu to the Apache-wide security page, making it more likely people will discover the Spark-specific one first.

Also adds a link from the Spark-specific security page to the security  page in the docs.
    
See also https://cwiki.apache.org/confluence/display/SECURITY/Documenting+your+security+model

### DIFF
--- a/_layouts/global.html
+++ b/_layouts/global.html
@@ -147,7 +147,6 @@
           <li><a class="dropdown-item"
                  href="https://www.apache.org/foundation/sponsorship.html">Sponsorship</a></li>
           <li><a class="dropdown-item" href="https://www.apache.org/foundation/thanks.html">Thanks</a></li>
-          <li><a class="dropdown-item" href="https://www.apache.org/security/">Security</a></li>
           <li><a class="dropdown-item" href="https://www.apache.org/events/current-event">Event</a></li>
         </ul>
       </li>
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -152,7 +152,6 @@
           <li><a class="dropdown-item"
                  href="https://www.apache.org/foundation/sponsorship.html">Sponsorship</a></li>
           <li><a class="dropdown-item" href="https://www.apache.org/foundation/thanks.html">Thanks</a></li>
-          <li><a class="dropdown-item" href="https://www.apache.org/security/">Security</a></li>
           <li><a class="dropdown-item" href="https://www.apache.org/events/current-event">Event</a></li>
         </ul>
       </li>
--- a/security.md
+++ b/security.md
@@ -7,6 +7,13 @@ navigation:
   show: true
 ---
 
+<h2>Security Model</h2>
+
+For information on what security properties to expect from Apache Spark
+and how to configure the various security features, see the
+[Spark Security](https://spark.apache.org/docs/latest/security.html)
+documentation.
+
 <h2>Reporting security issues</h2>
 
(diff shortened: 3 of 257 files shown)


srowen (approved):

srowen (changes requested):
Oh actually it looks like the generated HTML has some differences and the CI/CD checks fail. Can you check if your jekyll version, etc is the same? see the notes on https://github.com/apache/spark-website

raboof:
> Oh actually it looks like the generated HTML has some differences and the CI/CD checks fail. Can you check if your jekyll version, etc is the same? see the notes on https://github.com/apache/spark-website

ah I ignored some changes that seemed unrelated... now committed those as well

_layouts/global.html
```diff
@@ -147,7 +147,6 @@
           <li><a class="dropdown-item"
                  href="https://www.apache.org/foundation/sponsorship.html">Sponsorship</a></li>
           <li><a class="dropdown-item" href="https://www.apache.org/foundation/thanks.html">Thanks</a></li>
-          <li><a class="dropdown-item" href="https://www.apache.org/security/">Security</a></li>
```
yaooqinn, line 150: Instead of removing this, please update it with the  project-specific one and match the regex ` ^https?://.*apache.org/.*[Ss]ecurity` to satisfy ASF requirements

_layouts/home.html
```diff
@@ -152,7 +152,6 @@
           <li><a class="dropdown-item"
                  href="https://www.apache.org/foundation/sponsorship.html">Sponsorship</a></li>
           <li><a class="dropdown-item" href="https://www.apache.org/foundation/thanks.html">Thanks</a></li>
-          <li><a class="dropdown-item" href="https://www.apache.org/security/">Security</a></li>
```
yaooqinn, line 155: ditto

_layouts/global.html
srowen: I think the point is that there is already a "Security" link in the Developer menu that points to the Spark Security page.

yaooqinn: I mean we can replace this with our security side to reach the goal of this PR and satisfy the ASF website check service

raboof: I had assumed the Security link in the Developers menu would _already_ satisfy the ASF website check - but indeed I see the ASF website check (https://whimsy.apache.org/site/check/security) seems to want an absolute link to an apache.org site, while the Security link in the Developers menu is a relative link.

I guess we could fix this either on the spark-website side by replacing `{{site.baseurl}}` with `{{site.url}}{{site.baseurl}}` (see e.g. https://mademistakes.com/mastering-jekyll/site-url-baseurl/), or try to change the regex to also allow relative links.

yaooqinn: https://spark.apache.org/security.html can be used to satisfy both

raboof: Whimsy should [now](https://github.com/apache/whimsy/pull/270) accept relative links as well, so no further changes required on the Spark side AFAICS

yaooqinn:
https://github.com/apache/spark-website/actions/runs/14347401294/job/40219690897 

hi @raboof, can you send a followup to fix the CI?

raboof:
> https://github.com/apache/spark-website/actions/runs/14347401294/job/40219690897
> 
> hi @raboof, can you send a followup to fix the CI?

#604 should do it I think

PR #653, opened 2025-12-12
Document sbt build issues with IDEs

gaogaotiantian:
Added troubleshooting tips for using sbt with IDEs.

### DIFF
--- a/developer-tools.md
+++ b/developer-tools.md
@@ -494,6 +494,27 @@ in the Eclipse install directory. Increase the following setting as needed:
 256M
 ```
 
+<h4>IDE with sbt</h4>
+
+When you use IDEs like VSCode or Cursor, you need to build from terminal with `sbt`.
+A common issue you'll encounter is something like
+
+```
+java.lang.RuntimeException: Failed to find name hashes for
+```
+
+This is because multiple instances are trying to build the project simultaneously. Normally
+the culprit is your Java/Scala plugin in your IDEs.
+
+In order to fix this, first you need to disable auto-build of your plugins. Then you need to
+clear all the cache. Potential locations are
+
+```
+**/target/**
+~/.m2
+~/.ivy2/
+```
+
 <a name="nightly-builds"></a>
 <h3>Nightly builds</h3>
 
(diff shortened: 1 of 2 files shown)


developer-tools.md
```diff
@@ -494,6 +494,27 @@ in the Eclipse install directory. Increase the following setting as needed:
 256M
 ```
 
+<h4>IDE with sbt</h4>
+
+When you use IDEs like VSCode or Cursor, you need to build from terminal with `sbt`.
```
szehon-ho: should it be:  when you use ... and you build from terminal ?

Otherwise it sounds like if user use IDE => user needs to build using sbt

gaogaotiantian: Yeah that's true.

developer-tools.md
```diff
@@ -494,6 +494,27 @@ in the Eclipse install directory. Increase the following setting as needed:
...
+```
+
+This is because multiple instances are trying to build the project simultaneously. Normally
+the culprit is your Java/Scala plugin in your IDEs.
+
+In order to fix this, first you need to disable auto-build of your plugins. Then you need to
+clear all the cache. Potential locations are
+
+```
+**/target/**
```
cloud-fan: `sbt clean` can clear these?

developer-tools.md
```diff
@@ -494,6 +494,27 @@ in the Eclipse install directory. Increase the following setting as needed:
 256M
 ```
 
+<h4>IDE with sbt</h4>
+
+When you use IDEs like VSCode or Cursor and you build from terminal with `sbt`,
```
yaooqinn, line 499: Can we also provide tips for step-by-step setups for these IDEs? In the `IDE setup` section, we only have guidelines for `IntelliJ`. It's better to have instructions for them too

gaogaotiantian: To be honest I don't have that information now. I'm definitely not opposed to that idea but I don't have enough material at hand and the main purpose of this PR is to document this specific behavior. We can of course have a better write up about how to set VSCode/Cursor for spark project.

gaogaotiantian:
What was the CI failure about? I don't understand it.

gaogaotiantian:
I think I need an approval for the CI to run :)

cloud-fan (approved):

PR #660, opened 2026-01-11
Add a bit of more description for ASF_NEXUS_TOKEN

HyukjinKwon:
Per https://github.com/apache/spark/pull/53764

### DIFF
--- a/site/release-process.html
+++ b/site/release-process.html
@@ -636,7 +636,7 @@ and then send an e-mail to the mailing list with a subject that looks something 
   <li><code class="language-plaintext highlighter-rouge">ASF_PASSWORD</code>: The password for your ASF account.</li>
   <li><code class="language-plaintext highlighter-rouge">ASF_NEXUS_TOKEN</code>: ASF Nexus API token associated with your ASF account.
 Can be found in https://repository.apache.org/#profile;User%20Token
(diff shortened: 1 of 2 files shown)


site/release-process.html
```diff
@@ -635,7 +635,8 @@ <h2 id="preparing-your-github-actions-setup">Preparing your GitHub Actions setup
   <li><code class="language-plaintext highlighter-rouge">ASF_USERNAME</code>: Your Apache Software Foundation (ASF) account ID.</li>
   <li><code class="language-plaintext highlighter-rouge">ASF_PASSWORD</code>: The password for your ASF account.</li>
   <li><code class="language-plaintext highlighter-rouge">ASF_NEXUS_TOKEN</code>: ASF Nexus API token associated with your ASF account.
-Can be found in https://repository.apache.org/#profile;User%20Token</li>
+Can be found in https://repository.apache.org/#profile;User%20Token
+It is written in <code class="language-plaintext highlighter-rouge">&lt;password&gt;...&lt;/password&gt;</code>.</li>
```
dongjoon-hyun: It would be great if we can explicitly document that we ignore `User Code` part because we always `ASF_USERNAME` even when we use the generated token in the password part.

HyukjinKwon: yupyup!

dongjoon-hyun (approved):
+1, LGTM.

PR #682, opened 2026-03-23
[SPARK-55489] Update extended LTS period post SPIP ratification

holdenk:
Regenerated with docker container.
image


versioning-policy.md
```diff
@@ -115,4 +115,7 @@ For example, branch 2.3.x is no longer considered maintained as of September 201
...
 
 The last minor release within a major release will typically be maintained for longer as an "LTS" release.
-For example, 3.5.0 was released on September 13th 2023 and will be maintained for 31 months until April 12th 2026.
+For example, 3.5.0 was released on September 13th 2023 and would normally be maintained for 31 months until April 12th 2026.
+
+
+As an exception from the normal versioning policy, version 3.5.x has an "extended" LTS period to allow for migrations to be completed, as 4.0 was released relatively late into the 3.5 cycle. This extended LTS period will end *November 2027* . During the 3.5.x extended LTS period, we will only include security fixes. This extend LTS only applies to the primary Apache Spark project/repo and does not apply to sub projects with separate repos/releases (namely: Spark Connect for Swift/Rust/Go and Spark Kubernetes operator). Additionally, as Java 8 support is being removed from other projects (including Hadoop), should a dependency have a security fix that is not backported to a Java 8 compatible version we will drop Java 8 from the remainder of the extended LTS period.
```
dongjoon-hyun: Thank you, @holdenk .

BTW, this looks misleading to me. Shall we remove this statement simply to avoid any confusions in the community?
> as 4.0 was released relatively late into the 3.5 cycle.

holdenk: Sure we can drop that part.

holdenk: Done @dongjoon-hyun , look good?

versioning-policy.md
```diff
@@ -115,4 +115,7 @@ For example, branch 2.3.x is no longer considered maintained as of September 201
 of 2.3.0 in February 2018. No more 2.3.x releases should be expected after that point, even for bug fixes.
 
 The last minor release within a major release will typically be maintained for longer as an "LTS" release.
-For example, 3.5.0 was released on September 13th 2023 and will be maintained for 31 months until April 12th 2026.
+For example, 3.5.0 was released on September 13th 2023 and would normally be maintained for 31 months until April 12th 2026.
+
+
+As an exception from the normal versioning policy, version 3.5.x has an "extended" LTS period to allow for migrations to be completed. This extended LTS period will end *November 2027*. During the 3.5.x extended LTS period, we will only include security fixes. This extend LTS only applies to the primary Apache Spark project/repo and does not apply to sub projects with separate repos/releases (namely: Spark Connect for Swift/Rust/Go and Spark Kubernetes operator). Additionally, as Java 8 support is being removed from other projects (including Hadoop), should a dependency have a security fix that is not backported to a Java 8 compatible version we will drop Java 8 from the remainder of the extended LTS period.
```
pan3793: @holdenk, thank you for updating this, but it seems that the Java 8 dropping clause appears to be a new language that goes beyond what the community voted on.

I think no community member proposed or discussed a mechanism for dropping Java 8 mid-LTS if a dependency's security fix isn't backported to a Java 8-compatible version. On the contrary, Steve Loughran expressed:

> I think it's really good to have a stated security policy of maintenance,
> and that you have the engineering resources to do the release -including
> testing.
>
> I think you ought to make clear to all downstream that even security fixes
> are "best effort" and if compatibility issues surface you may not update a
> transient dependency with known CVEs.
>
> This is a real problem: what if there's an update to some library
> "random-12.0" which needs to be upgraded to "random-13.1" to fix a CVE, but
> that needs major code changes and could potentially break application code.
> What to do? Issue a potentially incompatible release or say "can't fix".

https://lists.apache.org/thread/4r6q8187b30p0ppclw0pyfjp8h8xs3rq

which is more in line with reality.

versioning-policy.md
```diff
@@ -115,4 +115,7 @@ For example, branch 2.3.x is no longer considered maintained as of September 201
 of 2.3.0 in February 2018. No more 2.3.x releases should be expected after that point, even for bug fixes.
 
 The last minor release within a major release will typically be maintained for longer as an "LTS" release.
-For example, 3.5.0 was released on September 13th 2023 and will be maintained for 31 months until April 12th 2026.
+For example, 3.5.0 was released on September 13th 2023 and would normally be maintained for 31 months until April 12th 2026.
+
+
```
pan3793: nit: one empty line

versioning-policy.md
holdenk: This was based on my reply to Steves comment in SPIP doc which where i said "It runs on 8/11/17. I agree in the end user description we’ll make it clear if there’s CVE in a sub library that doesn’t backport the fix to a API compatible Java 8 version we won’t be able fix the Java 8 version. Hopefully this doesn’t come up in practice but we shall see." Which Steve gave a "thumbs up" react to. I think no one voting on this assumed we would and backport fixes in other libraries. Is there alternative language you'd prefer here?

holdenk: Would "we may drop Java 8 or mark the issue as won't fix it" work for you? It doesn't commit us hard to one path or another.

dongjoon-hyun (approved):
+1, LGTM (from my side). Thank you, @holdenk !

versioning-policy.md
pan3793: I think "drop Java 8" in 3.5.x likely won't happen, for cases that 3rd party libs official release does not provide a Java 8 compatible security patched version, what is more likely to happen is that someone maintains a security fork if the issue is really serious, e.g., `log4j 1.x` => `reload4j`.

But I'm okay with the current words, thank you, @holdenk

pan3793 (approved):

versioning-policy.md
holdenk: Huzzah

PR #683, opened 2026-03-24
Document new security vulnerability in Apache Spark

miqowhy:
Added details about CVE-2025-55039 vulnerability in Apache Spark, including severity, affected versions, descriptions, mitigations, and credit.

Two remarks:
- I've taken the affected versions from @holdenk [description](https://lists.apache.org/thread/zrgyy9l85nm2c7vk36vr7bkyorg3w4qq), but I think all newer versions are also affected
- For @holdenk 2nd mitigation - I think setting spark.ssl.enabled=true is not enough, you also need spark.ssl.rpc.enabled=true, at least that's what the [documentation](https://spark.apache.org/docs/latest/security.html#ssl-configuration) says


holdenk:
Thanks for adding this! There's a build step you need to follow the for website changes to be visible.

miqowhy:
Thanks for the reply, I'll follow the build step.

But in parallel I think we need to clarify the affected versions - I'll be reproducing the CVE to check, but is there any reason why newer versions are not affected by this vulnerability? 

Looking at the docs of the latest version it looks like the vulnerable default is still in place. 

Or am I missing something?

security.md
```diff
@@ -58,6 +58,30 @@ This includes verifying the source, validating integrity, and applying appropria
 before loading or deploying a model.
 
 <h2>Known security issues</h2>
+<h3 id="CVE-2025-55039">CVE-2025-55039: Apache Spark: RPC encryption defaults to unauthenticated AES-CTR mode, enabling man-in-the-middle ciphertext modification attacks</h3>
```
pan3793, line 60: nit: a new line between <h2> <h3>

holdenk:
So I think https://issues.apache.org/jira/browse/SPARK-47172 indicates that the new config option is recommended for the new versions which if you follow the guidance in security settings is secure.

miqowhy:
> Thanks for adding this! There's a build step you need to follow the for website changes to be visible.

I've run the build step now, hope this is enough

miqowhy:
Anything preventing us from merging? @holdenk @pan3793

PR #688, opened 2026-05-14
Update release policy page

HyukjinKwon:
This PR updates Release Policy page per [SPIP: Accelerating Apache Spark Release Cadence](https://docs.google.com/document/d/1gBoZ4KH5zQUWpgK3M7zAN7p6Glz4S_e9bO3PvQA9sQs/edit?tab=t.0#heading=h.vj8hviw7ebqz).

### DIFF
--- a/site/versioning-policy.html
+++ b/site/versioning-policy.html
@@ -164,16 +164,20 @@ These small differences account for Spark&#8217;s nature as a multi-module proje
 <p>Each Spark release will be versioned: <code class="language-plaintext highlighter-rouge">[MAJOR].[FEATURE].[MAINTENANCE]</code></p>
 
 <ul>
-  <li><strong>MAJOR</strong>: Major releases occur annually, third-party dependency 
-upgrades, and major code refactoring. All releases with the same major version number will have 
-API compatibility.</li>
-  <li><strong>FEATURE</strong>: Feature releases occur quarterly (every 3 months) and contain new features, performance 
+  <li><strong>MAJOR</strong>: Major releases occur annually (every 12 months) as <strong>x.0.0</strong> releases. These releases may
+include <strong>breaking changes</strong>, third-party dependency upgrades, API removals and deprecations, and other changes that are not
+compatible with the previous major line. Within a major line, all <strong>x.y.z</strong> releases share API compatibility
+as described in the <strong>FEATURE</strong> and <strong>MAINTENANCE</strong> bullets below.</li>
+  <li><strong>FEATURE</strong>: Feature releases (<strong>x.y.0</strong> with <strong>y &#8805; 1</strong>) occur quarterly (every 3 months) and contain new features, performance 
 improvements, API additions, and bug fixes. To ensure safe and predictable upgrades for downstream 
 projects, feature releases have the following compatibility:
     <ul>
-      <li>No third-party dependency upgrades (e.g. Parquet, Arrow, ORC, Hadoop, Netty).</li>
-      <li>No behavior or semantic changes (e.g. SQL semantics, execution behavior, optimizer behavior, 
-configuration defaults). Exceptions might occur case by case (e.g., security issues).</li>
+      <li>No third-party dependency upgrades (e.g. Parquet, Arrow, ORC, Hadoop, Netty) by default. Upgrades
+required to address <strong>security</strong> issues may be allowed; any other exception is decided case by case by
+the release managers and the community.</li>
+      <li>No behavior or semantic changes (e.g. SQL semantics, execution behavior, configuration defaults).
+Optimizer changes that only improve performance while preserving query results are allowed; optimizer
+changes that alter results are not. Exceptions might occur case by case (e.g., security issues).</li>
       <li>Public APIs may be added but not changed or removed.</li>
     </ul>
   </li>
@@ -184,11 +188,9 @@ only fixes can be merged, then a final period where voting occurs on release can
 windows will be announced immediately after the previous feature release to give people plenty 
 of time.</p>
 <ul>
-  <li><strong>MAINTENANCE</strong>: Maintenance releases will occur on an ad hoc basis depending on specific patches 
-introduced (e.g. critical bug fixes and security patches) and their urgency. In general these releases 
-are designed to patch bugs. However, higher level libraries may introduce small features, such as a 
-new algorithm, provided they are entirely additive and isolated from existing code paths. Spark core 
-may not introduce any features.</li>
+  <li><strong>MAINTENANCE</strong>: Maintenance releases (<strong>x.y.z</strong> with <strong>z &#8805; 1</strong>) will occur on an ad hoc basis depending on specific patches
+introduced (e.g. critical bug fixes and security patches) and their urgency. They contain bug fixes and
+security patches only; they do not introduce new features.</li>
 </ul>
 
 <h3>Alpha components</h3>
@@ -286,8 +288,8 @@ available APIs.</p>
 
 <h2>Release cadence</h2>
 
-<p>Starting with Spark 4.3, feature releases occur quarterly (every 3 months), containing new features, 
-improvements, and bug fixes. Major releases occur annually (every 12 months), allowing breaking 
+<p>Starting with Apache Spark 4.3.0 (the first quarterly feature release after 4.2.0), feature releases occur quarterly (every 3 months), containing new features, 
+improvements, and bug fixes. Major releases occur annually (every 12 months), typically in a stable window once per year (subject to release discussion on the dev list), allowing breaking 
 changes and dependency upgrades. Maintenance releases happen as needed in between for critical 
 bug fixes and security patches.</p>
 
@@ -316,6 +318,24 @@ bug fixes and security patches.</p>
   </tbody>
 </table>
 
+<h3>Illustrative transition: 2026 and 2027</h3>
+
+<p>The calendar below is an <strong>example</strong> to show how the community expects to bootstrap the faster cadence;
+exact dates stay subject to the usual release-discussion and voting process.</p>
+
+<ul>
+  <li><strong>2026</strong>: Ship <strong>Apache Spark 4.2.0</strong> on the timeline above (the last <em>large</em> feature drop before the
+transition). <strong>4.2.x</strong> is not affected by this policy (development began before the SPIP); existing
+commitments, including the <strong>18</strong>-month maintenance window for <strong>4.2.x</strong>, are unchanged. After <strong>4.2.0</strong>
+is generally available, plan <strong>Apache Spark 4.3.0</strong> as the first quarterly feature release on the new
+train (for example, roughly <strong>three months</strong> after the <strong>4.2.0</strong> GA date for the start of the <strong>4.3</strong> merge/RC
+cycle&#8212;this is not a fixed rule, only an illustration of quarterly feature releases).</li>
+  <li><strong>2027</strong>: Ship <strong>Apache Spark 5.0.0</strong> as the next <strong>annual</strong> major. Follow with quarterly <strong>5.1.0</strong>,
+<strong>5.2.0</strong>, and <strong>5.3.0</strong> feature releases; <strong>5.3.0</strong> is the <strong>5.x</strong> LTS as the last <strong>5.x</strong> feature release
+(for example targeting calendar quarters <strong>2027 Q1</strong> through <strong>Q4</strong> if the <strong>5.0.0</strong> major lands early
+in the year).</li>
+</ul>
+
 <h2>Maintenance releases and EOL</h2>
 
 <p>The following table summarizes the maintenance window for each release type:</p>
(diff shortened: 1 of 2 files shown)


site/versioning-policy.html
```diff
@@ -164,19 +164,31 @@ <h3>Spark versions</h3>
 <p>Each Spark release will be versioned: <code class="language-plaintext highlighter-rouge">[MAJOR].[FEATURE].[MAINTENANCE]</code></p>
 
 <ul>
-  <li><strong>MAJOR</strong>: All releases with the same major version number will have API compatibility.
-Major version numbers will remain stable over long periods of time. For instance, 1.X.Y may last 
-1 year or more.</li>
-  <li><strong>FEATURE</strong>: Feature releases will typically contain new features, improvements, and bug fixes. 
-Each feature release will have a merge window where new patches can be merged, a QA window when 
+  <li><strong>MAJOR</strong>: Major releases occur annually, third-party dependency 
+upgrades, and major code refactoring. All releases with the same major version number will have
```
cloud-fan: `major code refactoring` does not need to wait for the yearly major release, we should mention breaking changes here instead.

site/versioning-policy.html
```diff
@@ -305,14 +318,48 @@ <h3>Spark 4.2 release window</h3>
...
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>Feature (x.y)</td>
+      <td>Every 3 months</td>
+      <td>6 months</td>
+    </tr>
+    <tr>
+      <td>LTS (final feature of a major)</td>
```
cloud-fan: shall we make it explicit that every `x.3` release is the LTS?

site/versioning-policy.html
```diff
@@ -164,19 +164,31 @@ <h3>Spark versions</h3>
...
-  <li><strong>MAINTENANCE</strong>: Maintenance releases will occur more frequently and depend on specific patches 
-introduced (e.g. bug fixes) and their urgency. In general these releases are designed to patch bugs. 
-However, higher level libraries may introduce small features, such as a new algorithm, provided 
-they are entirely additive and isolated from existing code paths. Spark core may not introduce 
-any features.</li>
+of time.</p>
+<ul>
+  <li><strong>MAINTENANCE</strong>: Maintenance releases will occur on an ad hoc basis depending on specific patches 
+introduced (e.g. critical bug fixes and security patches) and their urgency. In general these releases 
+are designed to patch bugs. However, higher level libraries may introduce small features, such as a
```
dongjoon-hyun: - First, what is the definition of `higher level libraries`? In the context, this could mean every module except `Spark Core`. In other words, `SQL`, `MLLIB`, `GRAPHX` and so on.

- Second, why do we need to allow this in the maintenance releases? I believe we had better keep the policy simple. In other words, no new features at all.
> However, higher level libraries may introduce small features, such as a new algorithm, provided they are entirely additive and isolated from existing code paths.

site/versioning-policy.html
dongjoon-hyun: Since this PR already mentions two exceptions like the following, it would be great to mention more specifically by giving examples like 5.3 or 6.3 in addition to `every x.3 release`.

> As an exception from the normal versioning policy, version 3.5.x has an ...

> For example, Spark 4.5 (the final 4.x feature release) would be maintained for 18 months from its release date.

dongjoon-hyun (commented):
The bootstrapping is always difficult. Given that, why don't we give an explicit example like 2026 and 2027 release plan, @HyukjinKwon and @cloud-fan ?

For example, I guessed like the following but it seems that this PR has a different timeline.
- 4.2.0 (The original plan): 2026 Summer + 18 months.
- 4.3.0 (LTS based on SPIP)
   - 4.3.0 RC1 will start after 3 months from the release day of Apache Spark 4.2.0 in 2026.
- In 2027, Spark 5 will start as the annual release with the quarterly feature releases.
   - Spark 5.0: 2027.1
   - Spark 5.1: 2027.4
   - Spark 5.2: 2027.7
   - Spark 5.3: 2027.10 (LTS based on SPIP)

I just want to understand our plan of migration toward this SPIP. So, if there is an explicit example for 2026 and 2027. It would be perfect for me and all. The above is only one of personal understanding.

dongjoon-hyun:
cc @peter-toth , too.

versioning-policy.md
```diff
@@ -110,11 +118,24 @@ in between feature releases. Major releases do not happen according to a fixed s
 
 <h2>Maintenance releases and EOL</h2>
 
-Feature release branches will, generally, be maintained with bug fix releases for a period of 18 months. 
-For example, branch 2.3.x is no longer considered maintained as of September 2019, 18 months after the release
-of 2.3.0 in February 2018. No more 2.3.x releases should be expected after that point, even for bug fixes.
+The following table summarizes the maintenance window for each release type:
+
+| Release Type | Cadence | Maintenance Window |
+| ----- | ----- | ----- |
```
peter-toth: Don't we need `Major` type with version `(x.0.0)` here?

versioning-policy.md
```diff
@@ -110,11 +118,24 @@ in between feature releases. Major releases do not happen according to a fixed s
...
+| Feature (x.y) | Every 3 months | 6 months |
```
peter-toth: Shall we use version `(x.1+.0)` for Feature and `(x.y.1+)` for Maintenance?

site/versioning-policy.html
```diff
@@ -164,19 +164,31 @@ <h3>Spark versions</h3>
...
-  <li><strong>FEATURE</strong>: Feature releases will typically contain new features, improvements, and bug fixes. 
-Each feature release will have a merge window where new patches can be merged, a QA window when 
+  <li><strong>MAJOR</strong>: Major releases occur annually, third-party dependency 
+upgrades, and major code refactoring. All releases with the same major version number will have 
+API compatibility.</li>
+  <li><strong>FEATURE</strong>: Feature releases occur quarterly (every 3 months) and contain new features, performance 
+improvements, API additions, and bug fixes. To ensure safe and predictable upgrades for downstream 
+projects, feature releases have the following compatibility:
+    <ul>
+      <li>No third-party dependency upgrades (e.g. Parquet, Arrow, ORC, Hadoop, Netty).</li>
```
holdenk: Excluding matching minor version or security issues yes?

HyukjinKwon:
I believe this is ready for another look.

site/versioning-policy.html
```diff
@@ -164,19 +164,32 @@ <h3>Spark versions</h3>
...
 
 <ul>
-  <li><strong>MAJOR</strong>: All releases with the same major version number will have API compatibility.
-Major version numbers will remain stable over long periods of time. For instance, 1.X.Y may last 
-1 year or more.</li>
-  <li><strong>FEATURE</strong>: Feature releases will typically contain new features, improvements, and bug fixes. 
-Each feature release will have a merge window where new patches can be merged, a QA window when 
+  <li><strong>MAJOR</strong>: Major releases occur annually (every 12 months) as <strong>x.0.0</strong> releases. These releases may
+include <strong>breaking changes</strong>, third-party dependency upgrades, and other changes that are not
+compatible within the same major line. Within a major line, all <strong>x.y.z</strong> releases share API compatibility
```
jingyimei: shoud be: other changes not compatible with the PREVIOUS major line?

jingyimei: do you actually mean: other changes that are not compatible with the PREVIOUS major line?

site/versioning-policy.html
```diff
@@ -164,19 +164,32 @@ <h3>Spark versions</h3>
...
+as described in the <strong>FEATURE</strong> and <strong>MAINTENANCE</strong> bullets below.</li>
+  <li><strong>FEATURE</strong>: Feature releases occur quarterly (every 3 months) and contain new features, performance 
+improvements, API additions, and bug fixes. To ensure safe and predictable upgrades for downstream
```
jingyimei, line 172: how about API deletion/deprecation? shall we mention in the Major block?

site/versioning-policy.html
```diff
@@ -164,19 +164,32 @@ <h3>Spark versions</h3>
...
+compatible within the same major line. Within a major line, all <strong>x.y.z</strong> releases share API compatibility
+as described in the <strong>FEATURE</strong> and <strong>MAINTENANCE</strong> bullets below.</li>
+  <li><strong>FEATURE</strong>: Feature releases occur quarterly (every 3 months) and contain new features, performance 
+improvements, API additions, and bug fixes. To ensure safe and predictable upgrades for downstream 
+projects, feature releases have the following compatibility:
+    <ul>
+      <li>No third-party dependency upgrades (e.g. Parquet, Arrow, ORC, Hadoop, Netty) by default. Upgrades
+required to address <strong>security</strong> issues may be allowed; any other exception is decided case by case by
+the release managers and the community.</li>
+      <li>No behavior or semantic changes (e.g. SQL semantics, execution behavior, optimizer behavior,
```
jingyimei: what optimizer behavior mean here? does optimizer change that brings perf improvement count for behavior change?

site/versioning-policy.html
```diff
@@ -164,19 +164,32 @@ <h3>Spark versions</h3>
...
-  <li><strong>MAJOR</strong>: All releases with the same major version number will have API compatibility.
-Major version numbers will remain stable over long periods of time. For instance, 1.X.Y may last 
-1 year or more.</li>
-  <li><strong>FEATURE</strong>: Feature releases will typically contain new features, improvements, and bug fixes. 
-Each feature release will have a merge window where new patches can be merged, a QA window when 
+  <li><strong>MAJOR</strong>: Major releases occur annually (every 12 months) as <strong>x.0.0</strong> releases. These releases may
+include <strong>breaking changes</strong>, third-party dependency upgrades, and other changes that are not
+compatible within the same major line. Within a major line, all <strong>x.y.z</strong> releases share API compatibility
+as described in the <strong>FEATURE</strong> and <strong>MAINTENANCE</strong> bullets below.</li>
+  <li><strong>FEATURE</strong>: Feature releases occur quarterly (every 3 months) and contain new features, performance
```
jingyimei: to be consistent with the other two blocks,  add version format like: Maintenance releases (<strong>x.y.z</strong> with <strong>z &#8805; 1</strong>)

site/versioning-policy.html
```diff
@@ -274,9 +287,10 @@ <h4>Alternatives to breaking an API</h4>
 
 <h2>Release cadence</h2>
 
-<p>The branch is cut every January and July, so feature (&#8220;minor&#8221;) releases occur about every 6 months in general.
-Hence, Spark 2.3.0 would generally be released about 6 months after 2.2.0. Maintenance releases happen as needed
-in between feature releases. Major releases do not happen according to a fixed schedule.</p>
+<p>Starting with Spark 4.3, feature releases occur quarterly (every 3 months), containing new features,
```
jingyimei: 4.2, or 4.3? because starting 4.2, in 3 months there will be 4.3.

site/versioning-policy.html
```diff
@@ -274,9 +287,10 @@ <h4>Alternatives to breaking an API</h4>
 
 <h2>Release cadence</h2>
 
-<p>The branch is cut every January and July, so feature (&#8220;minor&#8221;) releases occur about every 6 months in general.
-Hence, Spark 2.3.0 would generally be released about 6 months after 2.2.0. Maintenance releases happen as needed
-in between feature releases. Major releases do not happen according to a fixed schedule.</p>
+<p>Starting with Spark 4.3, feature releases occur quarterly (every 3 months), containing new features, 
+improvements, and bug fixes. Major releases occur annually (every 12 months), allowing breaking
```
jingyimei: for Major release, given it's anually, do you want to propose a window in a year, like May-ish, Sep-ish, etc?

site/versioning-policy.html
```diff
@@ -303,16 +317,78 @@ <h3>Spark 4.2 release window</h3>
...
+      <td>N/A (patches only)</td>
+    </tr>
+  </tbody>
+</table>
+
+<p>Release branches <strong>other than LTS</strong> will, generally, be maintained with bug fix releases for a period of
+6 months (see the <strong>Major</strong> and <strong>Feature</strong> rows in the table above).</p>
+
+<p>Under the quarterly cadence, the <strong>third</strong> feature release after each major&#8212;<strong>x.3.0</strong>&#8212;is designated
+as the LTS (Long-Term Support) release for that major line and is maintained for <strong>18 months</strong> (for example
```
jingyimei: instead of saying the 3rd release, is it safer to say the last feature release within a major line is the LTS, so that there are more flexibilities?

site/versioning-policy.html
HyukjinKwon: deprecation is fine but deletion is not. Let me explicitly mention removals

HyukjinKwon:
Should be ready for a look.

site/versioning-policy.html
```diff
@@ -303,16 +318,78 @@ <h3>Spark 4.2 release window</h3>
...
 </table>
 
+<h3>Illustrative transition: 2026 and 2027</h3>
+
+<p>The calendar below is an <strong>example</strong> to show how the community expects to bootstrap the faster cadence;
+exact dates stay subject to the usual release-discussion and voting process.</p>
+
+<ul>
+  <li><strong>2026</strong>: Ship <strong>Apache Spark 4.2.0</strong> on the timeline above (the last <em>large</em> feature drop before the
+transition). Treat <strong>4.2.x</strong> like other non-LTS branches (6 month maintenance window). After <strong>4.2.0</strong>
```
dongjoon-hyun: This is different from what we discussed, @HyukjinKwon . We agreed that Apache Spark 4.2.x is beyond the policy because it started before SPIP. So, it has 18-months maintenance window.
> Treat <strong>4.2.x</strong> like other non-LTS branches (6 month maintenance window).

Here, in the discussion, Spark 4.2.x is not supposed to be affected. (https://lists.apache.org/thread/t45w2soc18j1grrfrckcg7fcfdtdx09p)

> Absolutely agreed - this SPIP does not retroactively modify any existing commitments. 
> Spark 4.2.x must also follow the current policy because development has already progressed under that assumption.

In the vote, SPIP targets only 4.3.0+ like this.

> I would like to start a vote on the SPIP: Accelerating Apache Spark Release Cadence Target timing of this is 4.3.0 so 4.2.0 would be released as it was before.

dongjoon-hyun (changes requested):

dongjoon-hyun:
Thank you for updating, @HyukjinKwon .

site/versioning-policy.html
HyukjinKwon: Oh yeah. let me update it.

dongjoon-hyun (approved):
+1, LGTM. Thank you, @HyukjinKwon !

peter-toth (approved):

versioning-policy.md
cloud-fan: The wording here is a bit misleading. LTS is also annually, but it's 3 month after the latest feature release.

Strictly speaking, Feature release is also not 3 month cadency, as major release and LTS are not feature releases.

Maybe we can say it more general: we plan to ship a release every 3 months. We bump the major release version every 4 releases. The first release in the line is the major release, the last release in the line is LTS.

cloud-fan (approved):

HyukjinKwon:
Merged to asf-site.

PR #698, opened 2026-07-11
Revert "Add workflow to automatically generate and commit site HTML (…

nchammas:
…#697)"

This reverts commit 59719624d8f4bb548e1042654e60cb787b2824cf, which appears to have broken the site.

PR #699, opened 2026-07-12
Automatically generate and commit site HTML

nchammas:
Add workflows to automatically build the site and push the generated HTML. This eliminates the need for contributors to commit or review HTML changes.

## What's Changing

Workflow changes:
- `html-build.yml`: Add a new workflow that runs on PRs and simply generates the docs. This confirms the Jekyll build is not obviously broken.
- `html-push.yml`: Add a new workflow that runs on new commits to `asf-site`. It generates the docs and pushes the generated HTML automatically as a new commit to `asf-site`.
- `doc_gen.yml`: Delete this workflow since it is subsumed by the new workflows.

Website source changes:
- Stabilize the sitemap so it's generated in a deterministic manner, eliminating repeated diff noise.

Repo documentation changes:
- Update the README and release guide to account for the new workflows.
- Unify the build instructions so there are only two places to look: `committers.md` and the main repo README.
- Add important note about `content/` so its critical role is documented clearly for everyone.

This is a second attempt at #697 (which was reverted in #698).

## Why it's Changing

The current workflow creates a lot of friction. Every website source change generates corresponding HTML output changes. Both need to be committed and part of the PR. This makes the diff messy and difficult to review, especially if the touched source affects multiple pages.

Generating and committing the HTML is also just mechanical grunt work that can and should simply be automated.

## Future Work

There are two main improvements we can further make to the workflow here.
1. Split source from HTML output into two separate branches: source on `master`, HTML on `asf-site`.
2. Automatically publish a staging site for each PR.

These improvements are orthogonal to the work in this PR. We can tackle them next.


nchammas:
cc @cloud-fan @holdenk 

If you'd like to setup a staging site to test this out, I'm happy to help with that.

Otherwise, if/when we merge this, let's confirm the deployed production site is still working as expected so we can revert quickly if necessary.

holdenk:
I am -1 on this change until the in flight releases are complete and distributed I don't think it's worth the risk. After those are done (hopefully middle of this month but you never know), happy to revisit. To be clear this is an explicit but temporary veto on changing the website build until after 4.2.0/4.1.3/4.0.4/3.5.9 are all released. I think this is a change we do really need (and greatly appreciate @nchammas for doing the work here), just the timing doesn't work in my mind.

nchammas:
Yes, I agree, no rush. Let's wait until all these upcoming releases are complete.

Unless we invest in setting up a proper staging site (or per-PR staging sites, though it's [not clear][1] to me whether that's possible), the only way to fully test this new setup is with a production deployment.

[1]: https://github.com/apache/infrastructure-asfyaml/tree/76d241ccef02e5397e10c173ebf04c07525311ea#autostage

pan3793:
@nchammas, thank you for working on this area, I agree that the current spark-website repo is a disaster ... some ideas and experience from maintaining other ASF project website repos might be helpful:

1. use `master` branch to track only the source code (if we want to keep the history, might need to use the git command to rewrite the whole history and filter out the compiled output files)
2. leave the `asf-site` branch to track the compiled output (keep it orphaned, no need to track history to keep the size under control)
3. when pushing code to `master`, trigger a workflow to build and push the output to the `asf-site`

if we agree with this direction, then:

1. Spark PMC needs to ask the infra team to change the default branch to `master`
2. @nchammas you can debug that on your forked repo first, and compare the differences between the auto-generated `asf-site` (your forked repo) and the current `asf-site` (apache upstream)

nchammas:
@pan3793 - This was exactly [my original suggestion](https://github.com/apache/spark-website/pull/696#issuecomment-4773921749). However, @cloud-fan wanted to take a more conservative approach so as not to completely change the current workflow that committers are familiar with. I am happy to help with either approach.

Setting up a staging site (or multiple staging sites, one per branch) is also a good idea and orthogonal to splitting the source from the HTML output. Happy to help with that, too.

pan3793:
@nchammas, thanks for the info, I wasn't aware of that.

@cloud-fan I don't think tracking the compiled output files in the source tree makes sense, for example, we don't track `*.class` files for a Java/Scala project, can we try with two branches?
- `master` for source code only;
- `asf-site` for generated content, keep history at the beginning to allow us to restore in case something goes wrong, and eventually only keep the last commit to reduce the repo size

cloud-fan:
The commit history is already mixed with md and html files, so I don't think it's a must-have to separate the commit history now. It's orthogonal to the html auto generation.

nchammas:
Switching PR to draft until all the current releases are out, per [the earlier discussion](https://github.com/apache/spark-website/pull/699#issuecomment-4952351614).

nchammas:
> To be clear this is an explicit but temporary veto on changing the website build until after 4.2.0/4.1.3/4.0.4/3.5.9 are all released.

It looks like [these releases are all complete][1]. @holdenk - Am I OK to mark this PR as ready for review?

[1]: https://spark.apache.org/news/index.html

To summarize the next steps given what we've discussed so far:

1. Review and merge this PR. It will eliminate the need for anyone to generate or commit HTML in their PRs.
2. The committer who merges this PR in should check the site immediately after the build runs to confirm the site deployed correctly and no revert is needed. (I believe that what broke the site in #697 was not related to the generated HTML, but I cannot be sure since we don't have a staging site!)
3. In the future, we should explore the following further improvements to the website workflow, both of which are valuable:
    - Setup a dedicated staging site where we can test sensitive changes. If possible, setup automatic staging sites that spin up for each PR.
    - Split the Markdown source and HTML output into separate branches.

Does this sound good to everyone?

holdenk:
I think we're good we can remove my soft -1 now that the patch releases are done. I do want @celestehorgan to take a look since she ran into some difficulty when she went to update the website.

.github/actions/build-html/action.yml
```diff
@@ -0,0 +1,22 @@
...
+    required: false
+    default: "3.4"
+
+runs:
+  using: composite
+  steps:
+  - name: Set up Ruby and Bundler
+    uses: ruby/setup-ruby@v1
+    with:
+      ruby-version: ${{ inputs.ruby-version }}
```
celestehorgan, line 17: Non-blocking question: is there a chance that people (accidentally) commit the Gemfile.lock and (accidentally) bump the ruby version?

README.md
```diff
@@ -21,29 +45,16 @@ the Markdown files in the Docker container.
...
 
-To mark a block of code in your Markdown to be syntax highlighted by `jekyll` during the 
-compile phase, use the following syntax:
+The website is deployed automatically by [ASF Infra][infra]. The deployment configuration is tracked by [.asf.yaml](./.asf.yaml) and is [documented here][asf-docs].
 
-    {% highlight scala %}
-    // Your Scala code goes here, you can replace Scala with many other
-    // supported languages too.
-    {% endhighlight %}
+One deployment detail that appears to be critical is the presence of the [`content`](./content/) symlink to `site/`. Even though ASF Infra is [aware of Jekyll][jek], we perhaps do not have the exact setup required for them to automatically use our [Jekyll config](./_config.yml) to understand where the site content lives. Without the `content` symlink, the website will just show a plain directory listing of the files in this repo.
```
celestehorgan, line 52: Non-blocking: 

I saw that in commit fd110d9c7b9148750889148f29243d52e7b5cd03 you removed the `content` symlink and then readded it in d7a321672e79d854659934b6caa3784d199f61ee .. what exactly happened here?

I think it's actually worth _removing_ this note. What folks don't know can't hurt em' 😅  and I think you might actually confuse/send people down the wrong path otherwise.

celestehorgan (commented):
Some non-blocking comments, but thank you for doing this work!

.github/actions/build-html/action.yml
nchammas: Gemfile.lock is [already committed](https://github.com/apache/spark-website/blob/asf-site/Gemfile.lock). This is so that the Jekyll build uses a stable set of dependencies.

The Ruby version is set by the default on L9, unless overridden by the workflow calling this action. It would only change by a deliberate update to the workflows.

README.md
nchammas: The background here is that in #697 I attempted to remove this symlink because there was no documentation anywhere about it -- neither here nor on ASF Infra's website -- and it seemed to be a vestigial leftover [from 10 years ago](https://github.com/apache/spark-website/commit/fff0f57013b88c7b3f2906564421323f357e6d01). As you can see from the commit message, even when it was added the author was not sure it was needed.

When we merged #697 it broke the site. We are not 100% sure the removal of `content/` was the cause, but it's the most likely explanation. In the absence of a proper staging site, we cannot be sure until we merge this new PR here.

The purpose of this note is specifically to capture this learning experience so it's not repeated. Assuming my analysis is correct, not knowing the purpose of `content/` _did_ hurt us.

nchammas:
I'm ready to move forward with this. And since I can commit to this repo now, I can monitor the deployment and immediately revert this change if anything goes wrong.

1. Are there any outstanding concerns before I move forward?
2. I noticed that on this repo we sometimes use [GitHub's merge button][1] rather than the merge script. Is it OK if I use GitHub's "Squash and merge"? Do we need the merge script at all on this repo?

[1]: https://github.com/apache/spark-website/pulls?q=is%3Apr+is%3Amerged

nchammas:
@cloud-fan @holdenk @pan3793 - I'd like the go-ahead from at least one of you before I merge this PR. Do you have any outstanding concerns?

nchammas:
We're coming up on the 4.3.0 release and I don't want to make @HeartSaVioR's life more difficult, so I don't mind holding off again on moving this forward if he asks for that.

At the same time, I really do want to get this done so we can all move on from futzing with generated HTML output. I can just merge this in using my new powers as committer, but as this is a shared workflow for everyone I would prefer to have some formal buy-in on this PR before I do that.

@cloud-fan @holdenk @pan3793 - If you could take another look at this PR I would really appreciate it!

HyukjinKwon (approved):

README.md
```diff
@@ -1,10 +1,26 @@
...
+This repository captures the main Apache Spark website located at https://spark.apache.org. The programming docs under https://spark.apache.org/docs/ are [in the main Spark repo][main], not here. They are built separately for each release of Spark and then copied to the website under the `docs/` directory.
 
-Building the site requires [Ruby 3](https://www.ruby-lang.org), [Jekyll](http://jekyllrb.com/docs), and
-[Rouge](https://github.com/rouge-ruby/rouge). The most reliable way to ensure a compatible environment
-is to use the official Docker build image from the Apache Spark repository.
+[main]: https://github.com/apache/spark/tree/master/docs#readme
+
+## Contributing
+
+To contribute changes, build and test the site locally, then submit a pull request with your changes. You only need to commit changes to the Markdown source. A [GitHub Actions workflow](.github/workflows/html-push.yml) will generate the corresponding HTML under `site/` and push it for you.
```
cloud-fan: **Non-blocking (P2):** `js/downloads.js`, `site/static/versions.json`, and other non-Markdown files are website source too, and the release guide asks contributors to commit them. Saying only Markdown needs to be committed can therefore produce incomplete PRs. Please say to commit all source changes while excluding only ordinary Jekyll-generated site output.

.github/pull_request_template.md
```diff
@@ -0,0 +1,5 @@
+<!--
+Include your source changes, but not the generated HTML. A GitHub workflow will automatically generate and push the HTML under `site/` in a follow-up commit if necessary.
```
cloud-fan: **Non-blocking (P2):** This blanket instruction also covers release PRs, but the release process explicitly commits separately built docs under `site/docs/<version>`; Jekyll preserves that tree and cannot recreate it. Please exclude only ordinary Jekyll-generated site HTML and call out the versioned release-doc exception.

cloud-fan (commented):
## Review summary

**PR tags:** `behavior-change`

I found two non-blocking inconsistencies in the new contribution guidance and one validation-policy question. The README needs to cover all website source types, and the pull-request template needs to preserve the separately built `site/docs/<version>` release exception. Please also confirm whether accepting non-fatal Jekyll stderr was an intentional change from the deleted workflow. The automated build-and-publish structure otherwise matches the stated scope, and I found no P0 or P1 blocker, but the unresolved warning-policy choice prevents an unconditional readiness conclusion.

## Findings

**2 total: 0 P0, 0 P1, 2 P2, 0 P3.**

### Non-blocking (P2)

- **Broaden the source-file guidance** — `README.md:9` — see inline.
- **Preserve the release-docs exception** — `.github/pull_request_template.md:2` — see inline.

## Decision challenges

### Confirm the Jekyll warning policy

The deleted `doc_gen` workflow captured Jekyll stderr and failed on messages outside its allowlist, while the shared replacement action only checks the `bundle exec jekyll build` exit status. Was dropping that warning gate intentional? If not, please preserve equivalent diagnostic filtering in the shared action so both the PR and publication workflows retain it.

cloud-fan:
I don't see a merge-method rule established by this PR's code or description. Please use the spark-website maintainer convention for the merge itself; someone with authority over that repository policy should confirm whether squash-and-merge is acceptable.

cloud-fan:
I reviewed the current changes and have two scoped items: please clarify whether dropping the prior Jekyll stderr/warning gate is intentional, and update the README to say contributors should commit all source changes while omitting only generated site output.

PR #535, opened 2024-07-19
Improve the navigational capability for release manager's guide

yaooqinn:
Goes like this:

https://github.com/user-attachments/assets/1081ce4a-bd5f-4993-ad1c-82679168c402

PR #700, opened 2026-07-13
Use AnchorJS to automatically generate heading anchors

nchammas:
Right now if you look at a page on the main website like [Useful Developer Tools][current], you'll notice that you cannot easily click on a heading and get an HTML anchor that will let you link directly to that section. On the other hand, if you look at the [programming docs][prog], you'll see that you _can_ get anchor links by mousing over a heading.

This discrepancy is because [the programming docs include AnchorJS][anch], whereas the main website does not.

This PR adds AnchorJS to the main docs so readers can easily link to headings there like they can on the programming docs:

This adds the [latest release of AnchorJS][rel], which is 5.0.0. The programming docs are [on AnchorJS 1.1.1][old], which is more than 11 years old. In a matching PR, I will update the programming docs to also use AnchorJS 5.0.0 so they are consistent.

I built the docs locally and confirmed that the anchors work as expected.

[current]: https://spark.apache.org/developer-tools.html
[prog]: https://spark.apache.org/docs/latest/
[anch]: https://github.com/apache/spark/blob/392dd8dc39b2d6bb3fb341901cb27d4d7103dc34/docs/_layouts/global.html#L199
[rel]: https://github.com/bryanbraun/anchorjs/releases
[old]: https://github.com/apache/spark/blob/44c931f006194a833f09517c9e35fb3cdf5852b1/docs/js/vendor/anchor.min.js#L2


nchammas:
Because this touches `global.html`, the diff on the generated HTML is going to be massive. For this reason I'm going to wait until #699 is merged before asking for this to be formally reviewed and merged in.

nchammas:
cc @yaooqinn. As an example, this work here combined with the `{:toc}` kramdown directive would eliminate the need to manually add section ids or a table of contents as you had to do in #535.

PR #702, opened 2026-07-15
Regenerate Spark 4.2.0 release notes

cloud-fan:
Rewrites the Spark 4.2.0 release-notes post, reconstructed from the `v4.1.0..v4.2.0` commit range on `branch-4.2` and public Apache JIRA metadata (issue type, fix/affected versions, priority, umbrella/Epic parents). The existing frontmatter is preserved; only the post body changes.

Highlights covered: geospatial types (GA), Change Data Capture + Auto CDC, Arrow-optimized Python UDFs on by default, the NEAREST BY ranking-join SPIP, DSv2 transaction management, SQL path-based name resolution, metric views, the modernized Web UI, and Java 25 support.

Note: only the Markdown source is changed here — the rendered `site/releases/spark-release-4-2-0.html` still needs regeneration via `.dev/build-docs.sh` (I could not build the `docs-builder` image in my environment). Happy to follow up with the built HTML, or a committer can regenerate it.

This pull request and its description were written by Isaac.


nchammas:
Ah nice, this addresses my question about the release notes [to the dev list](https://lists.apache.org/thread/1lrzn4pskkkoy1sh73wcxkxmjq7whvg7).

And looking at [the release notes folder](https://github.com/apache/spark-website/tree/asf-site/releases/_posts), it looks like the dashes version of the URL is the usual format, and 4.1.0 and 4.1.1 are the exception to the rule.

cloud-fan:
@nchammas A useful and safe improvement is to let the github action generate the html files and push the commit to the PR branch. Is that doable? I've asked @huaxingao to help regenerate the html files due to my local env issues.

nchammas:
Yes, very doable. I avoided doing that so that PR diffs don't get polluted with noise from the HTML, which would also make review more annoying. But if we prefer to have the HTML pushed to PRs vs. `asf-site`, we can do that. Let me know if that's what you'd like to do on #699.

cloud-fan:
> I avoided doing that so that PR diffs don't get polluted with noise from the HTML

It's already the case today, so making html autogen is already a net win.

nchammas:
True, the HTML auto-gen alone is still a net win. I definitely prefer the more ambitious goal of keeping HTML diffs out of PRs (and thus separated from source commits in `asf-site`'s future history), but I'm fine with whatever committers would like to do.

releases/_posts/2026-07-14-spark-release-4-2-0.md
```diff
@@ -11,8 +11,505 @@ meta:
...
+Arrow-optimized Python UDFs and Arrow-based PySpark IPC are now enabled by default, bringing faster, lower-overhead Python execution to every PySpark user.
+
+Data Source V2 gains transaction management, schema evolution for `INSERT`, richer operation metrics, and enhanced partition-stats filtering, and SQL adds a path-based name-resolution model (`SET PATH`, qualified names, `CURRENT_PATH()`) and metric views for semantic modeling.
+
+To download Apache Spark 4.2.0, please visit the [downloads](https://spark.apache.org/downloads.html) page. For [detailed changes](https://issues.apache.org/jira/projects/SPARK/versions/12356380), you can consult JIRA.
+
+* This will become a table of contents (this text will be scraped).
+{:toc}
+
+### Highlights
```
nchammas, line 31: Random question but: Do we want to make this "highlights" section a convention for `MAJOR` and `MINOR` releases going forward? i.e. Any release matching `*.*.0`.

We could add a simple linter for it and catch when a release is about to go out without this section. The linter could also ensure that the version in the release URL matches `A-B-C` and not `A.B.C.`. Happy to add that.

viirya (approved):

releases/_posts/2026-07-14-spark-release-4-2-0.md
huaxingao: Good idea. Let's do this as a follow-up so it doesn't block the release notes. Feel free to open a PR for it.

huaxingao:
Merging this PR. Let's follow up if there are more comments.

PR #703, opened 2026-07-16
Add linter for release notes

nchammas:
Add a linter that identifies common issues with release notes so we can catch them before making a release.

Those issues include:
- URL format is incorrect. This could be because of the date, the version, or something else.
- A non-maintenance release does not include a section with highlights.

I limited some of the checks to releases from 2025 onwards so as not to flag long past releases that really don't need fixing. The only recent releases that need adjusting are a couple from last year which have the wrong URL format.

The linter script is pure Python with no external dependencies. It includes some quick tests.

Here are some example failures:

```
ERROR: 2025-12-16-spark-release-4.1.0.md: version '4.1.0' is malformed. Valid example: 4-1-0
ERROR: 2026-01-09-spark-release-4.1.1.md: version '4.1.1' is malformed. Valid example: 4-1-0
ERROR: 2026-07-14-spark-release-4-2-0.md: major/minor release is missing a highlights heading. Add: ## Highlights
```

The first two are real and are addressed as part of this PR. This will break any links that point to the old URLs, but I think this is acceptable in order to make the URLs all consistent. The last error I induced just as a demonstration.

[As proposed here.](https://github.com/apache/spark-website/pull/702#discussion_r3589719642)


nchammas:
cc @huaxingao

nchammas:
This might be overkill, but the release notes are a highly visible part of every release. Let me know if you think this linter provides enough value to justify itself.

More generally, I would like to explore more standardized Markdown linting in a future PR, but these rules for release notes are specific enough that we need a custom solution.

PR #705, opened 2026-08-06
[MINOR] Fix indentation on versioning policy

nchammas:
If you look at the [versioning policy](https://spark.apache.org/versioning-policy.html) you'll see that one of the paragraphs breaks the indentation of the bulleted lists.

This PR just fixes the indentation so the paragraph is correctly grouped under the "FEATURE" bullet.


nchammas:
Going to leave this in draft until #699 is merged. Then we can use this as another test of the new workflows.

